PR #257, opened 2015-01-08
preventing panic in merge_delegate.go

stephensearles:
I was getting a panic in this code when not using any merge delegate. This fixes it, if this is a reasonable solution for your design.


tboerger:
As we already found out on IRC ```memberlist``` have not been updated.

PR #260, opened 2015-01-22
Use the new Travis Docker infrastructure and stuff

sethvargo:


sethvargo:
@ryanuber any idea why this is failing?

ryanuber:
@sethvargo it looks like maybe something on Travis changed. The `setup_test_subnet` script brings up a bunch of IP's on 127.0.0.x to run the tests, but this used to only be required on Mac. It doesn't look like the script has changed since forever ago, so the only thing I can think of is that the IP's were previously routable on Travis, and now they are not.

ryanuber:
Oh, or if this is now building in docker, that could very well be the cause.

sethvargo:
Ah yea. This is running in Travis' new Docker. Let me try disabling that.

ryanuber:
@sethvargo it looks like we made it past the networking thing after your changes. I just pushed to master to fix another bug, so if you rebase this the tests **should** start passing again.

sethvargo:
@ryanuber it still seems to be failing

ryanuber:
@sethvargo earlier you had removed `sudo: false`. I think we still need that since the network scenario in Docker won't allow our tests to run.

sethvargo:
Okay. Trying again.

sethvargo:
@ryanuber okay, idk what's going on here... I'm just going to close this out and maybe revisit it another day.

Issue #247, opened 2014-11-07
Using -iface does not work on Windows


armon:
Seems that the `-iface` does not work as intended on Windows

Downchuck:
It appears this issue can be closed.

PR #261, opened 2015-02-01
agent: interface binding fix for Windows

dgv:
Hi @armon, this fixes handling of -iface on Windows (#247), as you can see is totaly related with https://github.com/golang/go/issues/5395, so we need to use net.IPAddr by now, I already tested it on WinXP, Win7 (i386) and Linux (amd64) also. 

Kind regards


armon:
Thanks!

PR #288, opened 2015-04-27
IPC: Tags - DeleteTags is not handled properly.

Reregistered:
This fix actually deletes the key from the map if a key is specified in the DeleteTags list.


armon:
@Reregistered I don't think this is necessary. The way this method works is by computing the new set of tags, and then invoking `SetTags`. In this case, there is nothing to "delete", so much as just not copying the old tags.

Reregistered:
Yes of course, apologies.

PR #289, opened 2015-04-30
Add missing dot slash for the command line examples

charleswhchan:


armon:
That is only necessary if Serf is not in your $PATH.

Issue #295, opened 2015-06-22
Bad overflow handling in GetBindAddr


josharian:
at https://github.com/hashicorp/serf/blob/master/testutil/addr.go#L18

```go
var bindNum byte

// ...

	bindNum++
	if bindNum > 255 {
		bindNum = 10
	}
```

That should probably be `bindNum == 0`. Caught by a [proposed vet check](https://go-review.googlesource.com/#/c/11252/).

Easy fix for a new contributor? :)

slackpad:
@josharian indeed that's a little shady! We'd be happy to merge a PR that fixes it :-)

ryanuber:
So I think this is actually intentional. Since the unit tests keep binding to the next address, once we reach 256 we need to start over again. It's not the absolute best because we could in theory try binding the same IP twice, but so far it hasn't been problematic because we don't have 250+ unit tests running concurrently, so they just get recycled. I think cycling through 10-255 is probably OK for now. A better fix might be to increment the ports we bind to and just stick with 127.0.0.1 for the bind address.

ryanuber:
@slackpad clued me in to the `byte` thing - Should have read the link. Makes sense!

PR #296, opened 2015-06-22
https://github.com/hashicorp/serf/issues/295

dansouza:
fixed overflow handling reported on https://github.com/hashicorp/serf/issues/295


slackpad:
Thanks!

Issue #267, opened 2015-02-17
send query response over TCP when response > QueryResponseSizeLimit


pmenglund:
To support large query responses (> `QueryResponseSizeLimit`) the response could be sent back using the TCP connection.

ryanuber:
@pmenglund makes sense to me! Thanks!

ryanuber:
Original thread was here: https://groups.google.com/forum/#!topic/serfdom/rqiFFNIOOhI

nanliu:
+1

pmenglund:
Thanks @nanliu :)

victorcoder:
+1

sehyoc:
+1

voodoo144:
+1

slackpad:
#342 helps make this better by making it configurable, though it's not the full TCP solution.

PR #299, opened 2015-08-13
make query size configurable

pmenglund:
since #267 hasn't been implemented during the last 6 months, I'm submitting this as a bandaid until it is


serf/serf.go
```diff
@@ -77,11 +77,13 @@ type Serf struct {
...
-	queryBroadcasts *memberlist.TransmitLimitedQueue
-	queryBuffer     []*queries
-	queryMinTime    LamportTime
-	queryResponse   map[LamportTime]*QueryResponse
-	queryLock       sync.RWMutex
+	queryBroadcasts        *memberlist.TransmitLimitedQueue
+	queryBuffer            []*queries
+	queryMinTime           LamportTime
+	queryResponse          map[LamportTime]*QueryResponse
+	queryResponseSizeLimit int // Maximum bytes size for response
```
slackpad: I'm not sure this code will build as-is. This should be added to the Config structure, not the Serf object itself.

pmenglund: ok, I'll change that

pmenglund:
Bah! This got messed up - I'll open a new PR which will look cleaner...

PR #300, opened 2015-08-14
Updates docs

charleswhchan:
* Reword explanation for `-discover`
* Add default port to -bind (matches -rpc-addr)


armon:
Thanks!

PR #304, opened 2015-08-31
Bumps protocol version back down as we've made memberlist smarter.

slackpad:
Over on https://github.com/hashicorp/memberlist/pull/50 we made memberlist smarter about interoperating with v2/v3 so we don't need to bump the Serf protocol; this should ease rolling out newer versions of Serf and Consul.


slackpad:
Tracking the build failure on #307 - it's not related to this change.

Issue #305, opened 2015-08-31
Support SRV records for cluster discovery


dalehamel:
Currently the 'discover' option of serf supports mDNS, but this depends on multicast, and thus is not a viable solution in AWS and other unicast-only environments.

Our current work-around for this is to implement 'anchor nodes' for our cluster, which bootstrap the cluster and have SRV records associated with them.

When other nodes come up, they find the anchor nodes by SRV record and join their cluster.

This would be quite easy to add natively to serf, and I'd be fine with drawing up the PR myself if you think it's something that might get accepted.

IMHO, supported a form of service discovering that's very reliable and proven (DNS SRV) for environments that need it (can't use mDNS) is quite worthwhile.

slackpad:
This seems like something we could pull in as long as it doesn't add a bunch of complexity (which I agree it shouldn't) or new deps (but I think we already have everything you'd need).

PR #310, opened 2015-09-06
Initial attempt at SRV based cluster discovery

dalehamel:
@slackpad for review.

This is an initial attempt at supporting SRV-based discovery for https://github.com/hashicorp/serf/issues/305.

If the code looks familiar, it's because I blatantly stole it from the mdns discovery code.

I've done an initial test manually and things seem to work.

Note that this does not support recursive SRV records, which is not allowed by the RFC anyways.


dalehamel:
I've added a second commit that allows for multiple SRV records to be supplied, though I'm not sure if the correct convention would be to stick with how tags do it (define -tags multiple times to get a list), or if accepting a string that is a CSV is fine.

dalehamel:
I've updated again so that it uses ```AppendSliceValue``` to allow multiple records to be specified.

dalehamel:
I've done real-world test, and this replaces our previous approach of a crontab quite nicely.

dalehamel:
ping @ryanuber as you also seem like a good reviewer here

dalehamel:
ping

dalehamel:
sorry to keep nagging, any chance we can get some reviewers on this or chat about it in IRC?

slackpad:
@dalehamel I should be able to look at this tonight or tomorrow. Working through a backlog of things, but this is on the list :-)

dalehamel:
Great, thanks! Sorry for nagging again :)

dalehamel:
@slackpad friendly reminder to bump this again :)

command/agent/command.go
```diff
@@ -737,6 +754,7 @@ Options:
                            can be reloaded during later agent starts. This option
                            is incompatible with the '-tag' option and requires there
                            be no tags in the agent configuration file, if given.
+  -srvrecord               SRV record to discover peers. Can be specified multiple times
```
slackpad: Should add a period to the end here.

command/agent/srv.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// Start the background workers
+	go m.run()
+	return m, nil
+}
+
+// run is a long running goroutine that scans for new hosts periodically
+func (m *AgentSRV) run() {
+	hosts := make(chan *net.SRV)
```
slackpad: MDNS had some buffering here which might make this less back-n-forth-y since the SRV record will generally have a list.

command/agent/srv.go
```diff
@@ -0,0 +1,90 @@
...
+	poll := time.After(0)
+	var quiet <-chan time.Time
+	var join []string
+
+	for {
+		select {
+		case h := <-hosts:
+			// Format the host address
+			addr := fmt.Sprintf("%s:%d", h.Target, h.Port)
+
```
slackpad: The seen tracking seems worth it since calling join takes locks and calls through a bunch of code.

command/agent/srv.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+			if n > 0 {
+				m.logger.Printf("[INFO] agent.srv: Joined %d hosts", n)
+			}
+
+			join = nil
+
+		case <-poll:
+			poll = time.After(srvPollInterval)
+			go m.poll(hosts)
```
slackpad: If things got really slow could we start accumulating these goroutines? Can you make the goroutines time out or will they potentially block forever?

slackpad:
Looking good - I made an initial pass with some feedback.

dalehamel:
thanks @slackpad, addressing your comments now :heart:

command/agent/command.go
dalehamel: done

command/agent/srv.go
dalehamel: @slackpad my concern with seen is that an SRV record will update and we won't catch it because the backing host may be the same record, but a different host not yet in the cluster.

I would rather have the freshest data, which is why 'seen' worries me for long running agents.

Scenario: agentA get's nuked, but it's host is rebuilt using the same DNS record. It will never be added by any node because it's already seen.

command/agent/srv.go
dalehamel: I much prefer the loop too, it's way simpler.

:+1:

command/agent/srv.go
slackpad: I see. Wouldn't agentA join based on its own query of the SRV records, though?

dalehamel: Not if agentA is the only thing in the SRV record (it would be trying to join itself).

In our case, only a subset of the agents are in the SRV record, we call them 'anchor records'.

slackpad: Ok that makes sense. The other thing you could do is search `serf.Members()` before the `Join()` attempt. That way if you lose track of a node you will attempt to re-join it based on the SRV record.

dalehamel:
@slackpad I've implemented the single-loop approach and it's much simpler. I'm going to add the seen and members stuff you suggested, as I had previously been considering that as well.

dalehamel:
@slackpad I've made the changes you've suggested here and am ready for another review:

* It's done using a single loop, waking every 60 seconds to try and join again. The actual call to findSRV has a timeout of 30 seconds, so that goroutines can't pile up, and we don't block forever waiting on it.
* Members are checked on every run, and we only return SRV records for members **not** in the cluster. Join will be called only when there are new members to try.
 * Note - this can be done repeatedly if the SRV record is bad, and it's pointing to a bad host. This should be evident from the logs.

I've tested this locally and it seems to behave as expected:

* I threw a sleep into findSRV that was longer than the timeout interval, and it causes the timeout to happen, but a new goroutine gets spawned on the next iteration after sleeping
* I've created a dummy SRV record that points to localhost at two different ports, and once they have found each other they stop looking.

One case that I have not covered (that I just thought of) is while a node is still in a failed state, it might still be in the member list and not get re-joined. I'll have to look into what states are acceptable here.

I am not maintaining a separate 'seen' hash, as I think that's unnecessary extra complexity, as we should assume that the SRV record is good (or will soon be good, once the TTL rolls around).

command/agent/srv.go
```diff
@@ -0,0 +1,114 @@
...
+	return m, nil
+}
+
+// run is a long running goroutine that scans for new hosts periodically
+func (m *AgentSRV) run() {
+
+	for {
+		// Set up a channel and select so we can timeout if findSRV takes too long
+		// set channel to 1 so that goroutines can't pile-up
+		c := make(chan []string, 1)
```
slackpad: I don't think this will do what you expect since the channel is made inside the loop. Even if you hoist it up, I think this'll accumulate goroutines that are blocked trying to write to the channel. I'd just make a blocking call to `findSRV()` and let it do its thing since there's not a good way to "recall" it if it does timeout.

command/agent/srv.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// findSRV looks up the SRV records and returns a slice of all SRV records
+// that are not currently cluster members
+func (m *AgentSRV) findSRV() []string {
+	var hosts []string
+
+	// map the members so that we only do a single O(n) search through members
+	known_members := make(map[string]bool)
```
slackpad: Since `members` is already a map, I'd just look it up down below and check for aliveness, rather than making an alternate data structure.

command/agent/srv.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, v := range m.agent.Serf().Members() {
+		if v.Status.String() == "alive" {
```
slackpad: It would be cleaner to compare with the constant `StatusAlive`.

command/agent/srv.go
slackpad: Had an interesting conversation with @ryanuber about this as well. Would it make more sense to make this a startup option like `-join_srv` and `-retry_join_srv`, and you'd basically call this once or in a loop until it succeeds? If a new server is coming up, it will use this to join the cluster once it finds somebody alive in the SRV list to join to. This will save the ongoing polling because only the people who want to join will be polling, everybody else is good because they are part of the cluster.

command/agent/srv.go
dalehamel: members is not already a map, it's an array - which is the problem here. In order to save searching through the list over and over, I do a one time mapping (O(n)) to a map so that I can get O(1) lookups.

https://github.com/hashicorp/serf/blob/master/serf/serf.go#L696-L697

Unless I'm misunderstanding you, and there is a way to get this as a map?

```
// Members returns a point-in-time snapshot of the members of this cluster.
func (s *Serf) Members() []Member {
```

command/agent/srv.go
dalehamel: agreed, fixed.

command/agent/srv.go
dalehamel: Yeah I agree, I had considered that this might be happening (I'm super new at go), but thanks for confirming it. I'll drop it back to a simple loop, and add the startup option you suggest.

dalehamel:
@slackpad I've addressed your comments, and implemented the optional retry that you and @ryanuber suggested.

I'm not clear on what you meant about members being a map. Is it a map somewhere else that's accessible? If so, I agree that's simpler - but Members() returns a slice, so I have to map it for runtime optimization.

command/agent/srv.go
dalehamel: I see what you mean: ```members       map[string]*memberState```, though unfortunately members is not exported. It feels a bit silly to serialize it to a slice only to put it back to a map.

I suppose I could add a RawMembers function to avoid this, but I'm not familiar with a way to do something like a 'const' return in go (to prevent the caller from modifying the value). It scares me to do something like this, because members is clearly intended to be private, and while I only need read only access I don't know of a way to ensure this.

slackpad: You are right I was looking at the private member. Making a local map seems fine.

dalehamel: :+1: how's the rest of this looking then?

command/agent/command.go
```diff
@@ -76,6 +76,10 @@ func (c *Command) readConfig() *Config {
 	cmdFlags.Var((*AppendSliceValue)(&tags), "tag",
 		"tag pair, specified as key=value")
 	cmdFlags.StringVar(&cmdConfig.Discover, "discover", "", "mDNS discovery name")
+	cmdFlags.Var((*AppendSliceValue)(&cmdConfig.SRVRecords), "srvrecord",
```
slackpad: Thinking about this, would WAN flavors make sense, too?

I think I'd go with `join-srv` and `retry-join-srv` (and `-wan` variants) so it has good symmetry with the other `join` forms. These would go into different buckets but processed in the same way. I think you could create a single new function based on what you have and pass it the bucket of SRV records to try, and a pointer to a function that handles the LAN/WAN and retry details so we don't have to write a bunch of variants like the current `Command.retryJoin` and `Command.retryJoinWan` are.

If you did this you could respect the existing retry interval and max attempts variables, too.

command/agent/srv.go
```diff
@@ -0,0 +1,120 @@
...
+	} else {
+		m.logger.Printf("[INFO] Starting SRV for one-shot join.")
+		// A one-shot attempt to try to join the cluster via SRV
+		go m.joinSRV()
+	}
+	return m, nil
+}
+
+// run is a long running goroutine that scans for new hosts periodically
+func (m *AgentSRV) poll() {
```
slackpad: Kind of related to above, I think even this one stops once it has been able to join something.

command/agent/command.go
dalehamel: I don't really understand this comment, so I'm going to read through the code to figure out more about how -wan works.

dalehamel: Ok, i think i see what's going on here, I'll patch this up.

dalehamel: I understand the comment about join-srv VS retry-join-srv, but I don't see how I can reuse anything to do with the wan / lan profiles, because they don't seem to have anything to do with joining:

https://github.com/hashicorp/memberlist/blob/9631bcf2ffb1632b05d00d7a756707dd15d81b86/config.go#L188-L198

The profiles only seem to be concerned with the actual health checking of nodes. Could you demonstrate where join changes its behavior based on wan/lan? I don't really get it.

dalehamel:
@slackpad I believe I've addressed your comment on more closely mirroring join and retry-join.

Please let me know if this doesn't do what you meant with respect to wan/lan profiles, but I think what you meant was to use the existing values for join retry attempts / intervals.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+		c.logger.Printf("[INFO] agent: Will check SRV again in %v", config.RetryInterval)
+		time.Sleep(config.RetryInterval)
+	}
+}
+
+func (c *Command) joinSRV(agent *Agent, replay bool, srvrecords []string) (int, error) {
+	records := c.findSRV(agent, srvrecords)
+	// Attempt the join only if there are new records
+	if len(records) > 0 {
+		n, err := agent.Join(records, replay)
```
slackpad: This is totally what I was thinking. The one remaining WAN-related detail (if you think SRV records would be useful for WAN joining in addition to LAN joining) is to add WAN versions of these. Instead of copy-paste the LAN ones, if we could pass a function down to do the join ( then I think you could re-use all of this stuff. Basically you'd pass in a function that calls `agent.Join()` for the LAN version, and `agent.JoinWAN()` for the LAN version.

dalehamel: @slackpad it might be useful, but I think it could probably make it into a different PR as for now it's basically YAGNI, at least for us.

Aside from that, are we good to merge here?

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
 	if config.Discover != "" {
 		c.Ui.Info(fmt.Sprintf("  mDNS cluster: %s", config.Discover))
 	}
+
 	return ipc
 }
 
 // startupJoin is invoked to handle any joins specified to take place at start time
```
slackpad: This comment is stale.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+	for {
+		// Try to perform the join
+		n, err := c.joinSRV(agent, config.ReplayOnJoin, config.RetryJoinSRV)
+
+		if err != nil {
+			c.logger.Printf("[ERR] agent: Failed to join via SRV: %v", err)
+		}
+
+		if err == nil && n > 0 {
+			c.logger.Printf("[INFO] agent: Join completed. Synced with %d initial agents", n)
```
slackpad: Need a `return` here.

command/agent/command.go
slackpad: This comment is stale.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+	records := c.findSRV(agent, srvrecords)
+	// Attempt the join only if there are new records
+	if len(records) > 0 {
+		n, err := agent.Join(records, replay)
+
+		if err != nil {
+			return n, err
+		}
+
+		if n > 0 {
```
slackpad: You can probably delete this since the calling code will print the same number.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+			c.logger.Printf("[INFO] agent: Joined %d hosts via SRV", n)
+		}
+
```
slackpad: It doesn't look like `n` ever gets returned. It might simplify this to check if len(records) == 0 and bail, de-nest, and then return `n, nil` at the end (you can't do that now because it's scoped inside the `if`.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+// findSRV looks up the SRV records and returns a slice of all SRV records
+// that are not currently cluster members
+func (c *Command) findSRV(agent *Agent, srvrecords []string) []string {
+	var hosts []string
+
+	// map the members so that we only do a single O(n) search through members
+	known_members := make(map[string]bool)
+	for _, v := range agent.Serf().Members() {
+		if v.Status.String() == serf.SerfAlive.String() {
```
slackpad: No need for `String()` -> `if v.Status == serf.SerfAlive`.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+			known_members[member] = true
+		}
+	}
+
+	// Look up each SRV record and check if it's already in the cluster
+	for _, record := range srvrecords {
+		_, srvhosts, err := net.LookupSRV("", "", record)
+
+		if err != nil {
+			c.logger.Printf("[ERR] agent: Failed to poll for new SRV hosts: %v", err)
```
slackpad: I'd probably add a continue here, and maybe include the offending srv record in the error message.

command/agent/command.go
```diff
@@ -427,10 +432,130 @@ func (c *Command) startAgent(config *Config, agent *Agent,
...
+		for _, host := range srvhosts {
+			// Find its addresses, as it's the only unique ID we can rely on
+			ipaddrs, err := net.LookupIP(host.Target)
+
+			if err != nil {
+				c.logger.Printf("[ERR] agent: resolve SRV record %s to IP %v", host.Target, err)
+			}
+
+			// For each address the host has, check if it's already in the cluster
+			for _, ipaddr := range ipaddrs {
```
slackpad: Now that we are just doing this at start time, you can simplify this by getting rid of the host checking. There's no ongoing polling, so this will only be called when this guy doesn't know about anyone else. Might as well try to join them all, since that's what will happen anyway.

slackpad:
Made another pass and noted a few things. I think it's super close - only remaining bits are the comments from this round, some unit tests, and doc updates.

command/agent/command.go
dalehamel: I'm wondering if it would make sense to make the return conditional, to provide something like a 'max trys = -1' to just keep polling, even after the cluster has been joined.

We have a use case for continuous polling.

command/agent/config.go
```diff
@@ -132,6 +132,15 @@ type Config struct {
...
 
+	// SRVRecords is used look for other agents using DNS SRV records.
+	// When this is set, the agent will look up the SRV record
+	// and attempt to add any hosts it finds. You can specify multiple times
+	// to look up multiple SRV records.
+	JoinSRV []string `mapstructure:"join-srv"`
+
+	// Exactly like join-srv, only keep trying until a successful join happens.
+	RetryJoinSRV []string `mapstructure:"retry-join-srv"`
+
```
ryanuber: These should probably be underscores to follow the config file convention of all of the other options.

command/agent/command.go
slackpad: What's the use case for continuous polling? Once you join, the cluster takes care of membership and a new guy can bootstrap with one of the agents in the srv record.

dalehamel: That assumes that the SRV record points to the hosts in the existing cluster. If that changes, you could have a split where you have two clusters who will never know about each other.

I think I mentioned this concern in another comment.

Let's say you have 4 nodes, A,B,C,D. Only A and B are in the SRV record. C and D join them.

A and B get torn down, and a new SRV record is built with E and F. You now have two clusters, DC, EF. If DC keep checking the SRV record ,they will find EF. If they don't, you'll have a partition.

slackpad: I see. Sorry for my confusion I just want to understand how this will be used and how it differs from existing options.

In your example, what kind of process decides who gets added to the SRV record? If you always populated the SRV record with all the Consul servers, for example, it would always be sufficient for new nodes to join.

I hope I haven't yagni-d you with the join options but that's where I thought it was going :-) For the polling case I'd had a -discover_srv option that takes the list and calls your function forever in a separate go routine vs. -1 retry count. If you can use the new join options without adding the polling support I still think that's ideal, though.

ryanuber: I'd also second that the continuous polling is ideally not Serf's responsibility. I think the common use case will be the one-shot join, and leaning on the gossip protocol to keep the list of members converged. I'm not totally opposed but I think we should have a very solid and general use case in mind before adding the additional complexity.

dalehamel: Our specific use case, which I think is quite applicable:

* We manage a VPC in terraform
* The VPC has jump hosts (bastion hosts) in the public subnet, which are required to bootstrap anything in the private subnet
* We use terraform to create SRV records for the VPC that all serf nodes use to find th e cluster
* All other nodes in the VPC discover the cluster using this SRV record pointing to these bastion hosts

All this is done as a workaround for the fact that mDNS can't work in ec2, so SRV records to these bastion hosts are created.

If the bastion hosts disappear, which might well happen, then we will have a cluster partition since we are only registering the bastion nodes in SRV records.

Our workaround right now is we have a crontab that looks up the SRV record and just sends a serf join on every host, but that feels like a huge hack when serf could support this natively.

slackpad: For this case, how do you bootstrap a new VPC + bastion initially (if there are no servers in the private subnet yet)? If I understand you correctly, the continuous polling is only useful when there are existing servers in the private subnet to find a new bastion that was added and join it to the cluster, right? (And the init time SRV join is useful when the bastion is up and new private subnet servers come up and need to join).

dalehamel: @slackpad the bastions find each other through the SRV record once it's registered, then all other nodes find the bastions.

The continuous polling is useful for any other nodes (private or public doesn't really matter) to discover the cluster. The only case where I think it's totally necessary is when both bastions go away, which would result in the original cluster becoming 'orphaned'.

A workaround would be to have the nodes register themselves as part of the SRV record, but I'd prefer to not have the nodes updating the records and to just keep them in terraform as that's a lot more predictable.

Or we could just stick with the crontab approach, but I think that the continuous polling to prevent orphaned clusters is the simplest, most reliable approach. Under the current implementation, it never has to try a join if the cluster is already fully joined.

command/agent/command.go
dalehamel: These are two different types, MemberStatus and SerfState, as far as I can tell the integers have no guarantee of matching, but the strings do map to the same thing.

command/agent/command.go
dalehamel: Leaving this one untouched until we decide what to do about polling in above comment thread.

dalehamel:
@slackpad @ryanuber I addressed most of the comments here.

Maybe the best way forward would be to just drop the polling as you guys suggested, and work on a separate PR to add that so that we can at least get the SRV stuff into core.

For the time being, we can continue to use a cron record and will still benefit from having SRV support in serf core.

dalehamel:
@slackpad @ryanuber as mentioned in my previous comment, i've gone ahead and removed the polling and performed the simplifications requested.

Ready for another pass, as I believe all comments have been addressed.

command/agent/command.go
```diff
@@ -76,6 +77,10 @@ func (c *Command) readConfig() *Config {
 	cmdFlags.Var((*AppendSliceValue)(&tags), "tag",
 		"tag pair, specified as key=value")
 	cmdFlags.StringVar(&cmdConfig.Discover, "discover", "", "mDNS discovery name")
+	cmdFlags.Var((*AppendSliceValue)(&cmdConfig.JoinSRV), "join-srv",
```
slackpad: These options should be `join_srv` and `retry_join_srv` with underscores.

slackpad:
The latest round looks good - can you add some unit tests and update the Serf docs (they are under /website in the same repo)?

Agree it would be good to get the startup SRV support in for now, and then add the polling later in a follow-on PR if needed.

command/agent/command.go
slackpad: Oops I picked the wrong constant - there's a MemberStatus `StatusAlive` that should work instead of the string (if we end up adding this back).

dalehamel:
@slackpad will do, i'm not too familiar with unit testing in go but i'll try and reverse engineer what you guys have.

One thing i was considering was adding SRV support to the join command as well, I'll take a look at scoping that out.

dalehamel:
@slackpad I've hit a bit of a road block on the unit testing. 

I need to fake an SRV record in some way, but I'm not sure how to stub this out in an elegant way.

Do you have any advice on how to approach this? I'm not sure what the best way to stub a DNS resolver in golang would be.

slackpad:
Hi @dalehamel - check out Consul's DNS unit test, it does this, specifically the framework-y part:

https://github.com/hashicorp/consul/blob/master/command/agent/dns_test.go#L51-L72

And here's an example usage of it:

https://github.com/hashicorp/consul/blob/master/command/agent/dns_test.go#L322-L325

dalehamel:
great, thanks @slackpad i'll take a look.

F21:
Any update on this?

ssenaria:
Bump for an update.

jason-riddle:
?

joeyguerra:
This is super old. Is anyone actively maintaining this?

dnephin:
Sorry for the silence on this PR. We had a look at this as part of trying to clean up some older PRs. The functionality seems to make sense. Other than a rebase, I think we would need a test to cover the new functionality.

If anyone is still interested in this and is able to reabse and add a test we can keep it open. Otherwise we'll probably close this PR in a few months.

Issue #306, opened 2015-08-31
Way to query config data for running agents


dalehamel:
It would be great if serf responded to a USR signal, or had an RPC call, for dumping the running config of an agent (in JSON).

This would help with debugging, as you could see how multiple JSON files were merged together.

In my case, I want to implement a query that will ask other serf agents to respond with what their event handlers are, so that I know what queries I can issue. I would like to achieve this with another handler script, that would just dump the running config and parse it for the implemented event handlers.

I would be happy to help implement this if it's accepted as a worthwhile feature.

slackpad:
Seems like a good idea. This would probably make the most sense as an RPC call (and maybe output as part of the `info` CLI, or possibly a new one). Consul has a similar capability at /v1/agent/self. We'd definitely want to redact sensitive information like encryption secrets, and only report a safe subset of the configuration.

slackpad:
Going to close this out since there haven't been additional requests for this and you found a simple solution for your use case in #312.

PR #311, opened 2015-09-07
Initial attempt at dumping running config with 'getconfig' RPC command

dalehamel:
@slackpad for review.

As discussed in https://github.com/hashicorp/serf/issues/306, this adds support for dumping the running serf config with a new RPC command 'getconfig'.

I've only added one import, with is 'crypto/sha256' so that I can redact sensitive fields (an alternative would be to just always print "REDACTED" instead, but I noticed that vault uses this pattern of sha256-summing sensitive data, so I figured it was a decent pattern to stick with).

This is a fairly naive attempt, and I don't really do much error handling (any), but I figure there isn't too much that can go wrong with just JSON marshalling the running config. I haven't added any tests yet either. I admittedly used a bit of a hack in order to support redacting data of marshalling then immediately unmarshalling the running config in order to obtain a copy. After redacting, I then pretty-print the JSON with indents.

Possibly enhancements:

* Tests
* Error handling
* Better way of knowing what data to redact
* Support for multiple output formats (though, I can't see anyone *not* wanting JSON here).

Sample output:

```
serf getconfig
{
    "NodeName": "localhost",
    "Role": "",
    "Tags": {},
    "TagsFile": "",
    "BindAddr": "0.0.0.0",
    "AdvertiseAddr": "",
    "EncryptKey": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "KeyringFile": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "LogLevel": "INFO",
    "RPCAddr": "127.0.0.1:7373",
    "RPCAuthKey": "",
    "Protocol": 5,
    "ReplayOnJoin": false,
    "StartJoin": [],
    "EventHandlers": [],
    "Profile": "lan",
    "SnapshotPath": "",
    "LeaveOnTerm": false,
    "SkipLeaveOnInt": false,
    "Discover": "",
    "Interface": "",
    "ReconnectIntervalRaw": "",
    "ReconnectInterval": 0,
    "ReconnectTimeoutRaw": "",
    "ReconnectTimeout": 0,
    "TombstoneTimeoutRaw": "",
    "TombstoneTimeout": 0,
    "DisableNameResolution": false,
    "EnableSyslog": false,
    "SyslogFacility": "LOCAL0",
    "RetryJoin": [],
    "RetryMaxAttempts": 0,
    "RetryIntervalRaw": "",
    "RetryInterval": 30000000000,
    "RejoinAfterLeave": false,
    "StatsiteAddr": "",
    "StatsdAddr": ""
}
```

### DIFF
--- a/command/agent/agent.go
+++ b/command/agent/agent.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -428,4 +429,32 @@ func (a *Agent) Stats() map[string]map[string]string {
 		"tags":    local.Tags,
 	}
 	return output
+}
+
+func (a *Agent) GetConfig() string {
+	a.logger.Printf("[INFO] agent: Dumping config")
+
+	var configCopy *Config
+
+	config_data, _ := json.Marshal(*a.agentConf)
+
+	// Here we unmarshal the data we just dumped
+	// As a hack to copy the agent config
+	// So that we can redact any sensitive data here
+	_ = json.Unmarshal(config_data, &configCopy)
+
+	// Redact sensitive data on our copy of the config
+	configCopy.EncryptKey = Redact(configCopy.EncryptKey)
+	configCopy.KeyringFile = Redact(configCopy.KeyringFile)
+
+	// Now we marshall it again, this time prettier
+	json_config, _ := json.MarshalIndent(configCopy, "", "    ")
+
+	return string(json_config)
+}
+
+func Redact(data string) string {
+	h := sha256.New()
+	io.WriteString(h, data)
+	return fmt.Sprintf("%x", h.Sum(nil))
 }
--- a/client/const.go
+++ b/client/const.go
@@ -13,6 +13,7 @@ const (
 const (
 	handshakeCommand       = "handshake"
 	eventCommand           = "event"
+	getConfigCommand       = "getconfig"
 	forceLeaveCommand      = "force-leave"
 	joinCommand            = "join"
 	membersCommand         = "members"
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -365,6 +365,18 @@ func (c *RPCClient) Stats() (map[string]map[string]string, error) {
 	return resp, err
 }
 
+// Stats is used to get the running config data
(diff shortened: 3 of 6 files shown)


command/agent/agent.go
```diff
@@ -429,3 +430,31 @@ func (a *Agent) Stats() map[string]map[string]string {
...
+	configCopy.EncryptKey = Redact(configCopy.EncryptKey)
+	configCopy.KeyringFile = Redact(configCopy.KeyringFile)
+
+	// Now we marshall it again, this time prettier
+	json_config, _ := json.MarshalIndent(configCopy, "", "    ")
+
+	return string(json_config)
+}
+
+func Redact(data string) string {
```
dalehamel: Note, by sha256summing empty strings, an attacker would be able to see that there is no encryption key set. Then again, they probably have already figured this out, so I don't think it really matters. Technically though, this is an information leak as the sha256sum of an already empty string is probably well known.

dalehamel:
As I mentioned in the original issue, one of my use cases for this is to write a serf handler that will query the local serf agent and respond with the events it handles, IE ```serf query handlers``` so that I can easily see what queries / events I can issue, and expect to be handled / get a response.

dalehamel:
ping @ryanuber as you also seem like a good reviewer here

ryanuber:
I think that we can avoid some of this by supplementing the `info` command. It already contains a subset of the configuration and supports JSON output, and could be augmented if you feel there are additional fields that would be useful to see, e.g. an `event_handlers` array with the names of handled events.

Additionally, part of Serf's security model assumes that the RPC connection between the client (CLI in this case) and the serf agent is safe, which is why it listens on loopback only by default. Redacting the encryption key is likely not necessary, and it would actually make more sense if we just omit that, as key rotation might make the key in the configuration obsolete anyways. If you do need to get the encryption key, you should use the `serf keys -list` command, which will dump the encryption keys currently in use on the cluster.

dalehamel:
@ryanuber alright, I'll rework this to instead extend the existing ```info``` command

dalehamel:
Closing in favor of https://github.com/hashicorp/serf/pull/312

PR #326, opened 2015-10-12
serf: on force-leave, send MemberLeave message to EventCh

armon:
This was causing sporadic test failures in Consul and Nomad. It should also make peer removal more timely instead of waiting for the reconcile period.

/cc: @slackpad @dadgar


dadgar:
LGTM

mitchellh:
A rare serf change emerges.

PR #328, opened 2015-10-19
Remove unreachable error check

kopiczko:
I believe this is a relict of the past. Error variable doesn't mutate here since last `err != nil` check.


slackpad:
Indeed - that's probably cruft or a bad merge. Thanks!

PR #331, opened 2015-10-22
Update dist script

sethvargo:


Makefile
```diff
@@ -1,30 +1,57 @@
...
-	./scripts/website_run.sh
-
-web-push:
-	./scripts/website_push.sh
-
-.PHONY: all cov deps integ subnet test web web-push
+	@sh -c "'$(CURDIR)/scripts/setup_test_subnet'"
+
+# test runs the test suite and vets the code
+test: generate
```
ryanuber: We still need the `subnet` target here so tests pass on OSX

Makefile
```diff
@@ -1,30 +1,57 @@
...
+	go test $(TEST) $(TESTARGS) -timeout=30s -parallel=4
+
+# testinteg runs the integration test suite
+testinteg: generate
```
ryanuber: Same as above about test target

Makefile
```diff
@@ -1,30 +1,57 @@
...
+	INTEG_TESTS=yes go test $(TEST) $(TESTARGS)
+
+# testrace runs the race checker
+testrace: generate
```
ryanuber: Need subnet target

ryanuber:
Added a few comments but otherwise LGTM!

PR #334, opened 2015-10-27
Fixes up some things in the Makefile.

slackpad:
* Adds the suffix to the subnet script call.
 * Gets rid of unused integ test target.
 * Runs the tests in serial, otherwise we get port conflicts.

/cc @sethvargo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,15 +24,11 @@ dist: #bin
 # subnet sets up the require subnet for testing on darwin (osx) - you must run
 # this before running other tests if you are on osx.
 subnet:
-	@sh -c "'$(CURDIR)/scripts/setup_test_subnet'"
+	@sh -c "'$(CURDIR)/scripts/setup_test_subnet.sh'"
 
-# test runs the test suite and vets the code
+# test runs the test suite
 test: subnet generate
-	go test $(TEST) $(TESTARGS) -timeout=30s -parallel=4
-
-# testinteg runs the integration test suite
-testinteg: subnet generate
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,15 +24,11 @@ dist: #bin
...
 
-# test runs the test suite and vets the code
+# test runs the test suite
 test: subnet generate
-	go test $(TEST) $(TESTARGS) -timeout=30s -parallel=4
-
-# testinteg runs the integration test suite
-testinteg: subnet generate
-	INTEG_TESTS=yes go test $(TEST) $(TESTARGS)
+	go list $(TEST) | xargs -n1 go test $(TESTARGS)
```
sethvargo: @slackpad interesting - what is the difference here? 

```
go list ./... | xargs -n1 go test
go test ./...
```

Makefile
```diff
@@ -24,15 +24,11 @@ dist: #bin
 # subnet sets up the require subnet for testing on darwin (osx) - you must run
 # this before running other tests if you are on osx.
 subnet:
-	@sh -c "'$(CURDIR)/scripts/setup_test_subnet'"
+	@sh -c "'$(CURDIR)/scripts/setup_test_subnet.sh'"
```
sethvargo: oops :smile:

sethvargo:
LGTM - one question, mostly for my own learning :smile:

Makefile
slackpad: The `-n1` will pull the packages off one at a time and feed them to a separate `go test` command in serial. Otherwise, I think `go test` will run tests for each package in parallel according to max procs.

PR #335, opened 2015-10-27
Captures the protocol versions for a node when an update event fires.

slackpad:
Fixes an incidental bug that I found while investigating https://github.com/hashicorp/consul/issues/1346. This will fix that issue, but for the wrong reason. We should merge this, especially for the note about how this isn't a complete fix in case we depend on this in the future. I'll make the real fix for the issue above over in Consul.


ryanuber:
Makes sense, and LGTM!

PR #337, opened 2015-11-01
Support restarting serf from within event handler script

schlomo:
The KillMode default of control-group will actually kill all spawned processes together with serf when you do `servicectl restart serf`. In the case of dpkg this leads to a circular problem where dpkg is aborted before the deb postinst script can finish.

This change fixes the problem and also gives serf 5 seconds to exit the cluster before systemd will kill it off with a SIGKILL.

### DIFF
--- a/ops-misc/systemd.conf
+++ b/ops-misc/systemd.conf
@@ -11,6 +11,12 @@ ExecPreStart=/usr/bin/install -d -g root -o root /etc/serf/
 ExecStart=/usr/local/bin/serf agent -config-dir=/etc/serf/
 # Use SIGINT instead of SIGTERM so serf can depart the cluster.
 KillSignal=SIGINT
+# Kill only the serf process and leave spawned processes running.
+# Required to support restarting serf from within a script spwaned by an event handler
+KillMode=process
+# Give serf some time to leave the cluster before killing it off
+TimeoutStopSec=5
+
 # Restart on success, failure, and any emitted signals like HUP.
 Restart=always
 # Wait ten seconds before respawn attempts.


slackpad:
Hi @schlomo - thanks for the PR. This feels a little dangerous, though I can see why you'd want this; I certainly have :-) I'm a little worried about leaked processes - have you had good experience with this in production?

schlomo:
Good question - I can't tell you. I am not running serf in a data center setup but use it to automate desktops in a school (see https://github.com/jschule/ubuntu/tree/gh-pages/src/jts-serf) so that leaked processes are no issue for me.

What I can say is that when I implemented this change I could not see any leftover processes after an upgrade.

You write that you also need this feature (restarting serf from within an event handler) - how did you do it? What other solution would you suggest?

Do you have packages where this file is installed by default? If not then the risk should be low as users would have to manually install the systemd configuration into their system.

slackpad:
@schlomo I've mostly wanted this when doing ad-hoc testing. In production we use an ssh-based thing totally separate from Serf to restart Serf (though typically we don't make any changes to existing infrastructure, we roll new nodes and tear down old ones). I think you are right that there's little risk to merging this and having anybody pick it up, though I don't think we want to have less conservative default settings when people copy these examples. For example, if Serf fails to restart you'll be stuck having to scramble and find another way to get your nodes online.

I'm thinking it's best not to take this since Serf isn't specifically designed to restart itself. I hope that makes sense, and this PR will be searchable so people will hopefully find this recipe if they want to try it. Thanks and I hope this makes sense!

PR #338, opened 2015-11-02
Add interval to reachability command

saromanov:
Hi! I just added additional option to reachability command. This is ```interval``` flag. For example ```serf reachability -interval 5``` call reachability test every 5 seconds. If this feature will be merged, ```reachability``` command might be extended as output mean time of test or output worse and best time of tests.

### DIFF
--- a/command/reachability.go
+++ b/command/reachability.go
@@ -40,6 +40,7 @@ Options:
 
   -rpc-addr=127.0.0.1:7373  RPC address of the Serf agent.
   -rpc-auth=""              RPC auth token of the Serf agent.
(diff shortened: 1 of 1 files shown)


slackpad:
Hi @saromanov thanks for the PR! I think for this basic use case you could do `watch -n5 serf reachability` so I'd probably opt to keep this code simpler. Having stats here is an interesting idea, but I'm not sure how useful it would be because there is so much going to to propagate events and gather them back up. For now I think we should leave this as a simple command that looks for dead nodes but doesn't report a lot of timing details.

Issue #252, opened 2014-12-12
Configurable profiles


initzero:
It would be nice to be able to adjust all the knobs associated with a memberlist.Config beyond the defaults (DefaultLANConfig() et al.).

For now I'm just patching in my own profile in:

https://github.com/hashicorp/serf/blob/master/command/agent/command.go.

``` go
serfConfig := serf.DefaultConfig()
switch config.Profile {
case "lan":
    serfConfig.MemberlistConfig = memberlist.DefaultLANConfig()
case "wan":
    serfConfig.MemberlistConfig = memberlist.DefaultWANConfig()
case "local":
    serfConfig.MemberlistConfig = memberlist.DefaultLocalConfig()
case "custom":
    customConf := memberlist.DefaultLANConfig()
    customConf.ProbeInterval = time.Second * 10
    customConf.ProbeTimeout = time.Second * 5
    customConf.GossipInterval = time.Second * 20
    customConf.GossipNodes = 2
    customConf.TCPTimeout = time.Second * 5
    serfConfig.MemberlistConfig = ultraConf
default:
    c.Ui.Error(fmt.Sprintf("Unknown profile: %s", config.Profile))
    return nil
}
```

It would be awesome if there was a way to load a custom profile from something like:

my_profile.json

``` javascript
{
    "GossipInterval": 10,
    "TCPTimeout": 5
}
```

Maybe build off DefaultLANConfig(), parse the file, and override any fields specified in it.

PR #341, opened 2015-11-13
Memberlist config

saromanov:
Inspiered by #252. Add additional option to config `profile_path` for path of .json file with memberlist configuration options. For example:

```
{
   "profile": "lan",
   "profile_path: "memberlist.json"
}
```

In this example, we load  `lan` profile, but with some changes in memberlist.json

memberlist.json:

```
{
    "gossip_nodes": 7,
    "tcp_timeout": "8s"
}
```

### DIFF
--- a/command/agent/command.go
+++ b/command/agent/command.go
@@ -278,6 +278,46 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
 		return nil
 	}
 
+	if config.ProfilePath != "" {
+		path := config.ProfilePath
+		f, err := os.Open(path)
+		defer f.Close()
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error reading '%s': %s", path, err))
+			return nil
+		}
+		profileconfig, err := DecodeProfile(f)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("ProfilePath: %v", err))
+			return nil
+		}
+
+		if profileconfig.TCPTimeout != 0 {
+			serfConfig.MemberlistConfig.TCPTimeout = profileconfig.TCPTimeout
+		}
+		if profileconfig.IndirectChecks != 0 {
+			serfConfig.MemberlistConfig.IndirectChecks = profileconfig.IndirectChecks
+		}
+		if profileconfig.SuspicionMult != 0 {
+			serfConfig.MemberlistConfig.SuspicionMult = profileconfig.SuspicionMult
+		}
+		if profileconfig.PushPullInterval != 0 {
+			serfConfig.MemberlistConfig.PushPullInterval = profileconfig.PushPullInterval
+		}
+		if profileconfig.ProbeTimeout != 0 {
+			serfConfig.MemberlistConfig.ProbeTimeout = profileconfig.ProbeTimeout
+		}
+		if profileconfig.ProbeInterval != 0 {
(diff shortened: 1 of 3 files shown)


command/agent/command.go
```diff
@@ -278,6 +278,46 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
 		return nil
 	}
 
+	if config.ProfilePath != "" {
+		path := config.ProfilePath
+		f, err := os.Open(path)
+		defer f.Close()
```
slackpad: This should go after the error check (if there's an error then the file is not open).

command/agent/command.go
```diff
@@ -278,6 +278,46 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
...
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error reading '%s': %s", path, err))
+			return nil
+		}
+		profileconfig, err := DecodeProfile(f)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("ProfilePath: %v", err))
```
slackpad: This can probably have the same error as above, `c.Ui.Error(fmt.Sprintf("Error reading '%s': %s", path, err))`

command/agent/config.go
```diff
@@ -202,6 +205,32 @@ type Config struct {
 	StatsdAddr string `mapstructure:"statsd_addr"`
 }
 
+// Profile is a configuration for custom profile
+// For description of fields, see memberlist config
+// https://github.com/hashicorp/memberlist/blob/master/config.go
+type Profile struct {
+	TCPTimeoutRaw string        `mapstructure:"tcp_timeout"`
```
slackpad: This looks like it needs a `go fmt` run.

command/agent/command.go
```diff
@@ -278,6 +278,46 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
...
+			return nil
+		}
+
+		if profileconfig.TCPTimeout != 0 {
+			serfConfig.MemberlistConfig.TCPTimeout = profileconfig.TCPTimeout
+		}
+		if profileconfig.IndirectChecks != 0 {
+			serfConfig.MemberlistConfig.IndirectChecks = profileconfig.IndirectChecks
+		}
+		if profileconfig.SuspicionMult != 0 {
```
slackpad: Missed handling `RetransmitMult` in here.

command/agent/config_test.go
```diff
@@ -354,6 +354,17 @@ func TestDecodeConfig_unknownDirective(t *testing.T) {
 	}
 }
 
+func TestDecodeProfile(t *testing.T) {
+	input := `{"tcp_timeout: 10s"}`
+	profile, err := DecodeProfile(bytes.NewReader([]byte(input)))
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+	if profile.TCPTimeout != 10 * time.Second {
```
slackpad: Could you add some tests for the other parts of the profile as well?

slackpad:
Hi @saromanov this is looking great! I noted a few small things and I think this will be good to go. If you'd like, you can update the website as part of this PR (docs are under website/) or I'm happy to take care of that for you after we merge. Either way is fine.

jen20:
Hi @saromanov! I've revived this as hashicorp/serf#519, and addressed the feedback left by @slackpad. I'll close out this pull request in favour of that one. Thanks for contributing to Serf!

Issue #343, opened 2015-11-22
sudo in 'make subnet' repeated loop is annoying


benfleis:
When typing `make`, I didn't trust an unexpected `sudo` popup, so I Ctrl-C'd it. Only to find myself Ctrl-C'ing it, a lot. Turns out to be just under 256 times. This is avoidable by reworking the subnet setup scripts. Diff coming.

PR #344, opened 2015-11-22
modify sudo and ordering, use set -e fail early

benfleis:
Avoid calling sudo 253 times if the first one gets Ctrl-C'd.
Tell the user that sudo is prompting, and why.
Move the OS sanity check to the front, pre ping.

https://github.com/hashicorp/serf/issues/343


pmenglund:
:+1:

slackpad:
@benfleis thanks for the fix!

PR #373, opened 2016-02-02
coordinate: rename file to ensure that it's built only on test files

fsouza:
When building non-test binaries, Go excludes any files that ends with
the _test suffix. As the testing package declares some global flags
using during tests, it's important to not import it in non-test
binaries, otherwise you will end-up with unwanted flags (like
-test.bench, -test.benchtime and -test.run).

I found this issue while creating binaries using [Gizmo](https://github.com/NYTimes/gizmo).


slackpad:
Thanks for the PR - that makes sense!

PR #380, opened 2016-02-24
Vendor serf upstream packages

sean-:
Both `godep` and `govendor` files are present for the time being.


freeformz:
Why bother with godep at all if you are going to start using govendor? Or vice versa.

sean-:
Hello @freeformz. This seems to be an interesting topic these days, especially with the release of Go 1.6 and official `vendor/` support.  As long as the correct data ends up in `vendor/` and comes with the right metadata, we're reasonably semi-agnostic to the tool that gets the job done.  In this case, serf's vendor packages started with an import from `godep`, however there was some inspection and explicit manipulation done with `govendor` that was appealing so I left its dependency management file in place.  Did you have a specific concern or interest, or were you just curious as to why there were two metadata files for two different programs present?

freeformz:
I was curious as to why there were 2 tools being used. 

Disclaimer: I'm a maintainer of godep, trying to bring it (slowly) out of the stone age.

Long term I feel like we're doing a poor job of tooling in the Go community around this and my (and I believe the author's of the other vendoring tools) expectations are that a project would only use a single tool. I hope eventually we can on a single meta-data file across tools though an am working towards that end.

sean-:
@freeformz thanks for that background!  I actually recognized your name from some help you lent phinze regarding a problem he had and checked to see you were really involved with godep.  Thank you, on both accounts.

I think everyone's looking for a better experience with regards to dependencies, godep being a big part of that evolution.  With regards to a single metadata file, that is really encouraging to hear.  This is the wrong place to ask if godep is going to migrate to reuse `vendor/`, but I'd be interested in hearing about that over email - sean@ - (or even coffee if you're local) and wanted to talk vendor management for a bit.  In general, I think everyone agrees that introducing two sources of truth for metadata isn't prudent (nor should it be necessary).

GNUmakefile
```diff
@@ -29,45 +30,40 @@ subnet:
...
-		| sort -u \
-		| xargs go get -f -u -v -d
-	@if [ "$(GITBRANCH)" != "" ]; then git checkout -q $(GITBRANCH); else git checkout -q $(GITSHA); fi
-	@if [ `git rev-parse HEAD` != "$(GITSHA)" ]; then \
-		echo "ERROR: git checkout has drifted and we weren't able to correct it. Was $(GITBRANCH) ($(GITSHA))"; \
-		exit 1; \
-	fi
-	@echo "INFO: Currently on $(GITBRANCH) ($(GITSHA))"
+# updatedeps installs all the dependencies needed to test, build, and run
+updatedeps:: tools
```
slackpad: Is the double colon legit make-fu or just a typo?

GNUmakefile
```diff
@@ -29,45 +30,40 @@ subnet:
...
+	govendor list -no-status +vendor | xargs -n1 go get -u
+	govendor update +vendor
 
 # generate runs `go generate` to build the dynamically generated source files
```
slackpad: I couldn't find any actual generate stuff in Consul so this is worth checking and deleting if that's the case for Serf, too, especially since this is a parent of everything.

GNUmakefile
sean-: Legit make arcanum:  https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html#Double_002dColon

GNUmakefile
sean-: I couldn't find anything other than `miekg/dns` that was using generate.  I left the target just in case, but am happy to nuke it.  2min.

GNUmakefile
```diff
@@ -6,68 +6,59 @@ GITBRANCH:=$(shell git symbolic-ref --short HEAD 2>/dev/null)
...
 
-# generate runs `go generate` to build the dynamically generated source files
-generate:
-	find . -type f -name '.DS_Store' -delete
-	go generate ./...
+get-tools::
+	go get -u -v $(GOTOOLS)
 
-.PHONY: default bin cov deps dev dist subnet test testrace updatedeps generate
+.PHONY: default bin cov dev dist subnet test testrace tools updatedeps vet
```
slackpad: Missed a few of the new targets in here, otherwise looks good.

PR #388, opened 2016-04-07
website: Fix mime types for some assets on deploy

phinze:
Should fix occassional issues with application/octet-stream mime type
assets breaking things on the site.


slackpad:
Thanks!

PR #396, opened 2016-06-15
Correct number of Serf events in docs

BobWhitelock:
I assume this wasn't updated when new event types were added in the past.

Thanks


slackpad:
Thanks!

PR #401, opened 2016-08-08
F update vendor

sean-:
Update `vendor/` to the latest set of bits with the exception of `golang.org/x/sys/unix`.  Once `go 1.7` has been released a second pass can be made to update `x/sys/unix`.


slackpad:
LGTM

PR #402, opened 2016-08-09
Fixes intent queue overflow by storing intents per-node.

slackpad:
This fixes https://github.com/hashicorp/consul/issues/1062 by storing the intents per-node. This will buffer all intents for a 5 minute period and intents are now stored in a map indexed per-node (later intents of each type by Lamport time will overwrite old ones). This should provide a reasonable footprint of memory usage without throwing out recent intents, which is what leads to the infinite rebroadcasts.


dadgar:
LGTM

Issue #386, opened 2016-04-02
vendoring member list makes serf useless as a library.


james-lawrence:
take for the following code example:
```go
import (
       "log"

	"github.com/hashicorp/serf/serf"
	"github.com/hashicorp/serf/vendor/github.com/hashicorp/memberlist"
)
type filter struct{}

func (t filter) NotifyAlive(peer *memberlist.Node) error {
	log.Println(peer.Name, peer.Addr.String(), int(peer.Port))
	return nil
}

func NewClusterOptions(nodeID, addr string, port int) Options {
	events := make(chan serf.Event)
	config := serf.DefaultConfig()
	config.MemberlistConfig = memberlist.DefaultWANConfig()
	config.MemberlistConfig.Alive = filter{}
	config.MemberlistConfig.BindAddr = addr
	config.MemberlistConfig.BindPort = port

	if nodeID != "" {
		config.NodeName = nodeID
	}

	return Options{
		Config:  config,
		EventCh: events,
	}
}

// Options holds the options for the cluster.
type Options struct {
	*serf.Config
	EventCh chan serf.Event
}
```

is code fails to compile because you can't import a packages vendored package. but if you try to import memberlist as a normal package:
```go
import (
"github.com/hashicorp/memberlist"
)
```
go complains about mismatched types.

sean-:
This is earily similar to the long conversation happening on go-nuts:


>A tricky (impossible?) situation with common packages and vendoring - Google Groups

https://groups.google.com/forum/m/#!topic/golang-nuts/AnMr9NL6dtc

Thanks for bringing this to our attention. We'll get something figured out. Do you have this problem if you synchronize memberlist in vendor and your GOPATH?

james-lawrence:
>Do you have this problem if you synchronize memberlist in vendor and your GOPATH?

yes because its based on the path, removing memberlist from my path also does not resolve the issue as memberlist still can't be located.

one solution is to move memberlist to a vendored directory closer to where its used, i.e.) the commands package. once that is done the problem goes away. Basically any exposed types can't be vendered in serf.

slackpad:
@james-lawrence I updated memberlist to HEAD in https://github.com/hashicorp/serf/pull/387 - if you try the latest Serf now does that help your issue?

slackpad:
Sorry that still won't fix it - just ran a local test from @sean-. Sorry about that.

cbednarski:
[...]

PR #403, opened 2016-08-09
F serf cmd

sean-:
Move the `serf` binary from the root of `github.com/hashicorp/serf` to `github.com/hashicorp/serf/cmd/serf`.  Serf's primary role is that of a library and therefore it should Serf should be as easy to embed out of the box.  The `make` and `make dev` targets remain unchanged and function as they did previously.

Fixes: #386 

Unrelated but worth noting that during the testing of this PR `darwin/arm` appears to be broken, but that looks like it's due to me using `go1.7rc6` and like this platform is now excluded from the build.


vendor/vendor.json
```diff
@@ -81,12 +81,6 @@
 			"revisionTime": "2015-12-06T04:24:12Z"
 		},
 		{
-			"checksumSHA1": "pWIsEUE53NffjS1goRml43Cvleg=",
```
slackpad: What's up with this diff?

sean-: `serf` now lo longer depends on `memberlist` so that it can be used as a library.

slackpad: Oh I was expecting that we'd have only a vendor down there under cmd - is that not the case?

sean-:
OOB :shipit: from @slackpad

vendor/vendor.json
sean-: Aye... "fixed."  See #404 and the updated comment I made at the bottom of #386 .

PR #405, opened 2016-08-09
Use `make bin` instead of `make` for the travis install command

sean-:
This prevents travis from testing Serf twice, once during `make` and a second time during `make test`.  Found while looking at:

https://travis-ci.org/hashicorp/serf/builds/150837335

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ branches:
     - master
 
 install:
-  - make
+  - make bin
   - make get-tools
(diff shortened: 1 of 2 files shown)


sean-:
The build takes longer, but this is a better test that compiles each of our target platforms.  `darwin/arm` had likely been broken for a while without anyone noticing.  This will catch this in the future and run the tests only once (previously they were running twice).

PR #406, opened 2016-08-13
build: Add target for Illumos

jen20:
This commit adds `solaris` to the list of cross-compiled operating systems such that binaries capable of running on Illumos will be produced by `make bin`.

No other changes are necessary to allow building.


slackpad:
LGTM!

PR #412, opened 2016-09-20
Expose BroadcastTimeout configuration parameter

zbjornson:
(Per discussion in https://groups.google.com/forum/#!topic/serfdom/SgKQQ9LkDw0)

Exposes the broadcast timeout config param, which needs to be tuned for cases when tons of nodes join at the same time and the default 5 seconds is insufficient.

Tested by scaling from 0 to ~500 nodes as fast as possible (~90 seconds) a few times with a 30s broadcast timeout. Seems to resolve issue discussed in thread.

### DIFF
--- a/cmd/serf/command/agent/command.go
+++ b/cmd/serf/command/agent/command.go
@@ -27,6 +27,9 @@ const (
 
 	// minRetryInterval applies a lower bound to the join retry interval
 	minRetryInterval = time.Second
+
+	// minBroadcastTimeout applies a lower bound to the broadcast timeout interval
+	minBroadcastTimeout = time.Second
 )
 
 // Command is a Command implementation that runs a Serf agent.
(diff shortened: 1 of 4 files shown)


slackpad:
Hi @zbjornson this looks great. Could you please drop a note on here as well:

https://github.com/hashicorp/serf/blob/master/website/source/docs/agent/options.html.markdown

Thanks!

zbjornson:
Thanks for the prompt review. CI fixed and docs added.

cmd/serf/command/agent/command.go
```diff
@@ -156,6 +171,13 @@ func (c *Command) readConfig() *Config {
 		config.RetryInterval = minRetryInterval
 	}
 
+	// Check for sane broadcast timeout
+	if config.BroadcastTimeout < minBroadcastTimeout {
+		c.Ui.Output(fmt.Sprintf("Warning: 'BroadcastTimeout' is too low. Setting to %v",
+			config.BroadcastTimeout))
+		config.BroadcastTimeout = minBroadcastTimeout
```
slackpad: I think you need to move this line up to get the proper error message (this will show the old value).

zbjornson: Oops, I blindly followed "RetryInterval" above. OK if I fix that for RetryInterval as well?

slackpad: Yeah good catch :-)

slackpad:
Thanks again! Glad this fixed your startup issue.

PR #439, opened 2017-01-27
Relay query responses through a few randomly chosen nodes

kyhavlov:
Adds messageRelayType for sending query responses through multiple (defaults to 3 for now) randomly chosen nodes for robustness.

### DIFF
--- a/serf/event.go
+++ b/serf/event.go
@@ -107,7 +107,7 @@ type Query struct {
 	addr        []byte    // Address to respond to
 	port        uint16    // Port to respond to
 	deadline    time.Time // Must respond by this deadline
-	relayFactor int       // Number of duplicate responses to relay back to sender
+	relayFactor uint8     // Number of duplicate responses to relay back to sender
 	respLock    sync.Mutex
 }
 
@@ -179,7 +179,7 @@ func (q *Query) Respond(buf []byte) error {
 			return fmt.Errorf("relayed response exceeds limit of %d bytes", q.serf.config.QueryResponseSizeLimit)
 		}
 
-		relayMembers := kRandomMembers(q.relayFactor, members, func(m Member) bool {
+		relayMembers := kRandomMembers(int(q.relayFactor), members, func(m Member) bool {
 			return m.Status != StatusAlive || m.ProtocolMax < 5 || m.Name == q.serf.LocalMember().Name
 		})
 
--- a/serf/event.go
+++ b/serf/event.go
@@ -96,18 +96,19 @@ func (u UserEvent) String() string {
 	return fmt.Sprintf("user-event: %s", u.Name)
 }
 
-// Query is the struct used EventQuery type events
+// Query is the struct used by EventQuery type events
 type Query struct {
 	LTime   LamportTime
 	Name    string
 	Payload []byte
 
-	serf     *Serf
-	id       uint32    // ID is not exported, since it may change
-	addr     []byte    // Address to respond to
-	port     uint16    // Port to respond to
-	deadline time.Time // Must respond by this deadline
-	respLock sync.Mutex
+	serf        *Serf
+	id          uint32    // ID is not exported, since it may change
+	addr        []byte    // Address to respond to
+	port        uint16    // Port to respond to
+	deadline    time.Time // Must respond by this deadline
+	relayFactor int       // Number of duplicate responses to relay back to sender
+	respLock    sync.Mutex
 }
 
 func (q *Query) EventType() EventType {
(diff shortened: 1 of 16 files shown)


serf/messages.go
```diff
@@ -136,6 +138,23 @@ func encodeMessage(t messageType, msg interface{}) ([]byte, error) {
 	return buf.Bytes(), err
 }
 
+// encodeRelayMessage wraps a message in the messageRelayType, adding the length and
+// address of the end recipient to the front of the message
+func encodeRelayMessage(t messageType, addr net.UDPAddr, msg interface{}) ([]byte, error) {
+	buf := bytes.NewBuffer(nil)
```
slackpad: For future expansion it's probably better to use msgpack for the header, too, and just write a struct out here. You should be able to run one encoder and then the other to put the two messages back to back. Should work similarly on the read side as well.

serf/event.go
```diff
@@ -156,13 +158,53 @@ func (q *Query) Respond(buf []byte) error {
 		return fmt.Errorf("response exceeds limit of %d bytes", q.serf.config.QueryResponseSizeLimit)
 	}
 
-	// Send the response
```
slackpad: I think we still want to keep the direct response as well. This'll keep things working if there are no other members in the cluster that support the relay.

serf/event.go
```diff
@@ -156,13 +158,53 @@ func (q *Query) Respond(buf []byte) error {
...
-	addr := net.UDPAddr{IP: q.addr, Port: int(q.port)}
-	if err := q.serf.memberlist.SendTo(&addr, raw); err != nil {
-		return err
+	// Relay the response through up to QueryResponseRelayLimit nodes
+	members := q.serf.Members()
+	relayMembers := kRandomMembers(q.serf.config.QueryResponseRelayLimit, members, func(m Member) bool {
```
slackpad: I wonder if the relay factor should come from a field in the query itself? That way the requestor can set the importance of it, and queries from an originator that doesn't support dedup will naturally have this set to 0, which is kind of nice.

serf/event.go
```diff
@@ -156,13 +158,53 @@ func (q *Query) Respond(buf []byte) error {
...
+		return m.Status != StatusAlive && m.ProtocolMax >= 5
```
slackpad: The filter approach is super nice for the version thing!

serf/query.go
```diff
@@ -93,20 +93,26 @@ type QueryResponse struct {
 	// respCh is used to send a response from a node
 	respCh chan NodeResponse
 
+	// acks/responses are used to track the nodes that have sent an ack/response
+	acks      map[string]bool
```
slackpad: For this kind of `set` use of a map I'd use `map[string]struct{}`.

serf/serf.go
```diff
@@ -25,7 +25,7 @@ import (
 // version to memberlist below.
 const (
 	ProtocolVersionMin uint8 = 2
-	ProtocolVersionMax       = 4
+	ProtocolVersionMax       = 5
```
slackpad: We will need to double check and see if this affects Consul and makes it harder to upgrade. For Consul we are "speaking" 2 but we "understand" 3 or whatever it is. If this complicates the upgrade we may want to do similar here, since we are switching based on what the receiver understands, so it's safe to run in a mix.

serf/event.go
```diff
@@ -156,13 +158,53 @@ func (q *Query) Respond(buf []byte) error {
...
 
-	// Clera the deadline, response sent
+	// Clear the deadline, response sent
 	q.deadline = time.Time{}
 	return nil
 }
+
+// kRandomMembers selects up to k members from a given list, optionally
+// filtering by the given filterFunc
+func kRandomMembers(k int, members []Member, filterFunc func(Member) bool) []Member {
```
slackpad: I think this is fine as-is but it's interesting if you upgrade a few nodes of a large cluster I bet this'll come up empty handed sometimes. I think that's actually a good feature because you won't overwhelm a small amount of nodes with relays as you are upgrading things, and it'll get progressively better at finding nodes once there are a lot of them that support relay.

kyhavlov: We could ensure that this always returns at least the local node, and leave out the direct response above, I'm not sure which is simpler though.

serf/serf.go
kyhavlov: Can you elaborate on this? Nothing here should be backwards-incompatible, so it doesn't seem like this should affect Consul upgrades at all.

serf/event.go
slackpad: That would work, though it's probably clearer to have the direct response vs. asking yourself to relay, in case there's some way to lose that packet to yourself somewhere in the network stack.

serf/serf.go
slackpad: I'd have to dig through the code again to remember how this part works, but there's a similar mapping in Consul - https://github.com/hashicorp/consul/pull/2684.

That maps Consul protocol versions to Serf versions. We need to make sure we account for the 5 there properly (if at all).

kyhavlov: It looks consul is on the latest serf version already, so it seems like this shouldn't be an issue: https://github.com/hashicorp/consul/blob/master/consul/config.go#L41-L47

serf/messages.go
```diff
@@ -136,6 +138,35 @@ func encodeMessage(t messageType, msg interface{}) ([]byte, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	buf := bytes.NewBuffer(nil)
+	handle := codec.MsgpackHandle{}
+	encoder := codec.NewEncoder(buf, &handle)
+
+	buf.WriteByte(uint8(messageRelayType))
+	buf.WriteByte(uint8(len(headerBuf.Bytes())))
```
slackpad: I think you can encode the header as a struct w/o the size.

serf/serf.go
slackpad: One other thing to check is there anything to worry about with Serf itself. We might want to keep [this](https://github.com/hashicorp/serf/blob/master/serf/config.go#L243) set to 4. For consul I make a constant like `ProtocolVersion4Compatible = 4` to signal that we aren't taking the max.

serf/serf.go
```diff
@@ -1282,18 +1283,30 @@ func (s *Serf) handleQueryResponse(resp *messageQueryResponse) {
 
 	// Process each type of response
 	if resp.Ack() {
+		// Exit early if this is a duplicate ack
+		if _, ok := query.acks[resp.From]; ok {
```
slackpad: If we can, let's add a test that shows the dedup stuff working.

serf/serf.go
```diff
@@ -1242,14 +1242,15 @@ func (s *Serf) handleQuery(query *messageQuery) bool {
...
-			deadline: time.Now().Add(query.Timeout),
+			LTime:       query.LTime,
+			Name:        query.Name,
+			Payload:     query.Payload,
+			serf:        s,
+			id:          query.ID,
+			addr:        query.Addr,
+			port:        query.Port,
+			deadline:    time.Now().Add(query.Timeout),
+			relayFactor: s.config.QueryResponseRelayLimit,
```
slackpad: We should pull this out of config and make it come from the `QueryParam` struct - that way the query-er gets to pick what happens. We'd also need to add that to the `serf query` CLI and API client. This'll give users full control of this feature. For Consul we can do the same when we integrate this. For keyring ops we can maybe just set it based on the size of the cluster.

serf/messages.go
```diff
@@ -136,6 +140,27 @@ func encodeMessage(t messageType, msg interface{}) ([]byte, error) {
 	return buf.Bytes(), err
 }
 
+type relayHeader struct {
```
slackpad: I'd add a comment about what this is used for.

slackpad:
Latest changes look good. I think the only remaining thing is probably to keep https://github.com/hashicorp/serf/blob/master/serf/config.go#L243 set to 4. See https://github.com/hashicorp/serf/pull/439#discussion_r98311005.

serf/query.go
```diff
@@ -41,10 +45,12 @@ func (s *Serf) DefaultQueryTimeout() time.Duration {
 
 // DefaultQueryParam is used to return the default query parameters
 func (s *Serf) DefaultQueryParams() *QueryParam {
+	relayFactor := math.Ceil(math.Log10(float64(s.memberlist.NumMembers() + 1)))
```
slackpad: We might want to leave this at zero for now to be conservative for library users and let people opt-in. I think in Consul we could do this for the keyring operations.

serf/serf.go
kyhavlov: I'm still not sure why we would want to stop defaulting to the latest version here; there shouldn't be any changes that cause incompatibilities between clients running protocol v4 vs v5, so why keep the default at 4?

PR #449, opened 2017-02-08
Add support for directly specifying the logger instead of the log output

nathanielc:
This is basically the exact same as the memberlist package.


slackpad:
Thanks!

PR #450, opened 2017-02-08
Remove the execute flag from the created snapshot files

nathanielc:
Unless I am unaware of some odd feature there is no need to set the execute bit on the file permissions of created snapshot files.


slackpad:
Nope - this looks good. Thanks!

Issue #426, opened 2016-11-10
Race found on Serf.members


tyler:
Greetings.

During a test run we happened to notice the race detector fire for an interaction between memberlist and serf. (The full trace is reported below.) I believe the issue is this bit of code (and the other two instances directly below this one) at serf.go:325...

```go
	serf.broadcasts = &memberlist.TransmitLimitedQueue{
		NumNodes: func() int {
			return len(serf.members)
		},
		RetransmitMult: conf.MemberlistConfig.RetransmitMult,
	}
```

`serf.members` is supposed to be protected by the `serf.memberLock` mutex. But it appears that we're finding the length of it here without grabbing the mutex.

I'll happily put in a PR to fix it, but before I do that I just wanted to make sure this isn't already a known issue that for some reason doesn't need to be fixed.

Full trace:
```
==================
WARNING: DATA RACE
Write at 0x00c420132330 by goroutine 262:
  runtime.mapassign1()
      /go-workspace/go/src/runtime/hashmap.go:442 +0x0
  github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf.(*Serf).handleNodeJoin()
      /go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf/serf.go:871 +0xf9e
  github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf.(*eventDelegate).NotifyJoin()
      /go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf/event_delegate.go:12 +0x56
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).aliveNode()
      /go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:950 +0xc9d
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).mergeState()
      /go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:1115 +0x3b0
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).mergeRemoteState()
      /go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/net.go:1002 +0x6f5
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).handleConn()
[...]

PR #452, opened 2017-02-09
Fix race condition in NumNodes for TransmitLimitedQueue

tyler:
Fixes https://github.com/hashicorp/serf/issues/426 by using `serf.NumNodes` in place of the `len(serf.members)` calls in the instantiation of TransmitLimitedQueue.


serf/serf.go
```diff
@@ -751,6 +745,13 @@ func (s *Serf) Members() []Member {
 	return members
 }
 
+func (s *Serf) membersLength() int {
```
slackpad: Hi @tyler I knew this seemed familiar :-) There's an existing `NumNodes()` method on the `Serf` you can call that does this.

tyler: Oh jeez, how did I miss that... fix incoming.

PR #460, opened 2017-03-28
Fix a few races around the QueryResponse

nathanielc:
There were two races in the QueryResponse type:

1. In the Finished method the read on r.closed needs to be protected
2. When sending the response on the channel a check needs to be made if the channel has been closed while holding the lock.

Both of these races have been fixed with this PR.

### DIFF
--- a/serf/query.go
+++ b/serf/query.go
@@ -1,6 +1,7 @@
 package serf
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


nathanielc:
Any update here? I have been running off this PR for several weeks now. It would be nice to get it merged into master.

kyhavlov:
Hi @nathanielc, sorry for the delay and thank you for the PR!

Issue #435, opened 2016-12-30
Possible data race in accessing serf.eventJoinIgnore


ilog:
serf/delegate.go:233 is accessing the serf.eventJoinIgnore without acquiring the joinLock

WARNING: DATA RACE

Read at 0x00c4201822a0 by goroutine 39:
  github.com/hashicorp/serf/serf.(*delegate).MergeRemoteState()
      /go/src/github.com/hashicorp/serf/serf/delegate.go:233 +0x8e3
  github.com/hashicorp/memberlist.(*Memberlist).mergeRemoteState()
     /go/src/github.com/hashicorp/serf/vendor/github.com/hashicorp/memberlist/net.go:958 +0x803
  github.com/hashicorp/serf/vendor/github.com/hashicorp/memberlist.(*Memberlist).handleConn()
      /go/src/github.com/hashicorp/serf/vendor/github.com/hashicorp/memberlist/net.go:231 +0xc39

Previous write at 0x00c4201822a0 by goroutine 22:
  github.com/hashicorp/serf/serf.(*Serf).Join.func1()
      /go/src/github.com/hashicorp/serf/serf/serf.go:634 +0x3f
  github.com/hashicorp/serf/serf.(*Serf).Join()
      /go/src/github.com/hashicorp/serf/serf/serf.go:649 +0x145

deckarep:
https://github.com/hashicorp/serf/pull/464 is my attempt to fix this data-race.  

1. Although the Serf object properly locks the mutex internally: https://github.com/hashicorp/serf/blob/master/serf/serf.go#L595 for the scope of the function.
2. The `MergeRemoteState` method has an unsynchronized access to the `serf.eventJoinIgnore` field here: https://github.com/hashicorp/serf/blob/master/serf/delegate.go#L254

jen20:
This was addressed by 4f3b7b4 as per the comments on #464, so I'm going to close this issue out. Please feel free to re-open if there is further information here!

PR #464, opened 2017-04-22
Handles the panic for eventJoinIgnore data-race state.

deckarep:
### Summary
There exists an unsynchronized read within the `delegate.go` file where a data-race was identified by another user.

- [x] - closes https://github.com/hashicorp/serf/issues/435
- [x] - moves the `joinLock` mutex above the `eventJoinIgnore` field of which it protects.
- [x] - in the `Serf.Join` function moves to fine-grained locking to avoid a dead-lock scenario.
- [x] - exposes an exported function called: `EventJoinIgnore` which locks and returns a copy of the state of the boolean. 
- [x] - refactors the `delegate.MergeRemoteState` method to get the `eventJoinIgnore` bool state only through the newly exposed public method.

### DIFF
--- a/serf/delegate.go
+++ b/serf/delegate.go
@@ -248,10 +248,10 @@ func (d *delegate) MergeRemoteState(buf []byte, isJoin bool) {
 		d.serf.handleNodeJoinIntent(&join)
 	}
 
-	// If we are doing a join, and eventJoinIgnore is set
+	// If we are doing a join, and EventJoinIgnore() is set
 	// then we set the eventMinTime to the EventLTime. This
 	// prevents any of the incoming events from being processed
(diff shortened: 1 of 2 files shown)


deckarep:
This approach should be safer because it leaves the internal lock semantics within the implementation of the Serf object.

This is contrary to having the delegate attempt to lock the `joinLock`.

preetapan:
@magiconair I think your fix in https://github.com/hashicorp/serf/commit/4f3b7b437dfb9632a27e60c7edd74ea35ac4e8b4 addresses this so this PR can be closed. Can you take another look here in case there's more to do?

magiconair:
yes, this can be closed.

PR #472, opened 2017-06-13
Add a config option to disable writing keyring to a file

kyhavlov:
For hashicorp/consul#2630

### DIFF
--- a/serf/config.go
+++ b/serf/config.go
@@ -222,10 +222,6 @@ type Config struct {
 	// persist changes to the encryption keyring.
 	KeyringFile string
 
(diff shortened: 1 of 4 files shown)


serf/config.go
```diff
@@ -222,6 +222,10 @@ type Config struct {
 	// persist changes to the encryption keyring.
 	KeyringFile string
 
+	// DisableKeyringFile prevents serf from writing keyring data to the
```
slackpad: Could we use an empty `KeyringFile` name to indicate that this is disabled w/o a new config param?

PR #477, opened 2017-07-03
Snapshot write fail recover

preetapan:
This attempts a compaction every 30 seconds after an error when appending to the snapshot file (both local and remote). After this is merged we should be able to gracefully recover from disk availability or other intermittent write errors without needing a restart.

This fixes #477

### DIFF
--- a/serf/snapshot.go
+++ b/serf/snapshot.go
@@ -31,6 +31,7 @@ const flushInterval = 500 * time.Millisecond
 const clockUpdateInterval = 500 * time.Millisecond
 const coordinateUpdateInterval = 60 * time.Second
 const tmpExt = ".compact"
+const snapshotErrorRecoveryInterval = 30 * time.Second
 
 // Snapshotter is responsible for ingesting events and persisting
 // them to disk, and providing a recovery mechanism at start time.
@@ -55,6 +56,7 @@ type Snapshotter struct {
 	rejoinAfterLeave bool
 	shutdownCh       <-chan struct{}
 	waitCh           chan struct{}
+	lastAttemptedCompaction        time.Time
 }
 
 // PreviousNode is used to represent the previously known alive nodes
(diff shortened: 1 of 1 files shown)


serf/snapshot.go
```diff
@@ -320,7 +320,16 @@ func (s *Snapshotter) appendLine(l string) error {
 
 	n, err := s.buffered.WriteString(l)
 	if err != nil {
-		return err
+		wrErr := err
```
slackpad: This feels like it could be pretty aggressive if there are errors for other reasons and lots of writes are being attempted since it'll try to snapshot per line that Serf tries to write. I think we should throttle the recovery attempts to something like 30 seconds. It also might be good to put this logic up up `tryAppend()` since that's kind of focused on handling an append error already, this would be an additional clause after that to check to see if it's time to try to compact and log if that fails.

preetapan: I thought about that, but figured it should fail fast if its low on disk or other resources. Seems valid to only do this every 30 seconds or so though, so I'll fix this.

serf/snapshot.go
```diff
@@ -311,6 +313,18 @@ func (s *Snapshotter) processQuery(q *Query) {
 func (s *Snapshotter) tryAppend(l string) {
 	if err := s.appendLine(l); err != nil {
 		s.logger.Printf("[ERR] serf: Failed to update snapshot: %v", err)
+		now := time.Now()
+		if now.Sub(s.lastAttemptedCompaction) > snapshotErrorRecoveryInterval {
+			s.lastAttemptedCompaction = now
+			wrErr := err
+			s.logger.Printf("[ERR] serf: Attempting compaction to recover from error...")
```
slackpad: This could probably be `[INFO]` since it's not really another error.

serf/snapshot.go
```diff
@@ -311,6 +313,18 @@ func (s *Snapshotter) processQuery(q *Query) {
...
+			err = s.compact()
+			if err != nil {
+				s.logger.Printf("[ERR] serf: Failed compaction due to error %v, will reattempt in %v", err, snapshotErrorRecoveryInterval)
```
slackpad: Errors in Go often have multiple parts, so don't inline very well, I'd change this to:

`[ERR] serf: Compaction failed, will reattempt after %v: %v", snapshotErrorRecoveryInterval, err`

and drop the original error since it was already printed up on line 315:

`[INFO] serf: Finished compaction, successfully recovered from error state`

PR #478, opened 2017-07-07
Fix Serf.Stats() panic when coordinates are disabled

lorenzosaino:
If the Serf agent is configured with coordinates disabled (i.e. `serf.Config.DisabledCoordinates = true`), invoking `Serf.Stats()` causes a segmentation fault. That's because, in order to set the value of `"coordinate_resets"` in the stats map, `Serf.Stats()` invokes `s.coordClient.Stats()`, with `s.coordClient `being a nil pointer.

This pull request fixes this issue by setting the value of `"coordinate_resets"` in the stats map only if coordinates are enabled.

### DIFF
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -1655,18 +1655,20 @@ func (s *Serf) Stats() map[string]string {
 		return strconv.FormatUint(v, 10)
 	}
 	stats := map[string]string{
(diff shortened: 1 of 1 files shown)


slackpad:
Thanks!

Issue #487, opened 2017-10-16
Metric serf.coordinate.adjustment-ms negative


chemicL:
The **serf.coordinate.adjustment-ms** metric can sometimes stay at a negative value of -9223372013568.000000, which results in our statsd logs being flooded by warnings (it says DEBUG, although it's at warning level).

```
Oct 13 06:25:05 statsd1 nodejs[1049]: 13 Oct 06:25:05 - DEBUG: Bad line: -9223372013568.000000,ms in msg "our.stats.nodeX.serf.coordinate.adjustment-ms:-9223372013568.000000|ms"
Oct 13 06:25:05 statsd1 nodejs[1049]: 13 Oct 06:25:05 - DEBUG: Bad line: -9223372013568.000000,ms in msg "our.stats.nodeY.serf.coordinate.adjustment-ms:-9223372013568.000000|ms"
```

It seems negative adjustments could be valid, however it's surprising that from different machines it's the same negative value repeated in all the logs.

We're using consul version 0.7.5 - 0.9.3 in our environment and that's where we observed the issue. The metric has been present in serf used by those versions.

slackpad:
Hey @chemicL thanks for the report - I don't recognize this as any of the coordinate-related constants so will need to look through the algorithm a bit. When you say "stay" does it get stuck like that essentially forever?

slackpad:
Linking to https://github.com/hashicorp/consul/issues/3023, which might be related.

chemicL:
I confirm it does "get stuck like that essentially forever", producing gigabytes of metrics for that single metric per day. All with the same value.

The metric is reported here: https://github.com/hashicorp/serf/blob/master/serf/ping_delegate.go#L78

slackpad:
@chemicL if you have any nodes in that state can you do a quick check of the v1/agent/self output and see if `coordinate_resets` is non-zero? It would be interesting to see if that's going up as well.

chemicL:
This happens also in version 0.7.5, where I don't see `coordinate_resets` in the v1/agent/self output.
In an agent at 0.9.5 which reports the corrupt metric: `"coordinate_resets": "0"`

slackpad:
One final thing I just thought of on your 0.9.3 agent in the bad state can you snag the state of `Coord` inside /v1/agent/self? That might shed some light on what's happening. We don't have plans to release a 0.9.4 but if you don't mind building locally you could delete that log line from the ping delegate as a simple workaround.

[...]

PR #488, opened 2017-10-19
Protect against ping periods that are out of range

preetapan:
I could reliably reproduce both extreme negative adjustment values and negative rtt estimates by feeding it time.maxvalue as the rtt value, which is a valid return values from the `time.Sub` method. 

This patch makes it so that we reject any updates that contain rtt durations that are not in a range from 1 to 32400 seconds (9 hours). Another option is to force updates > 9 hrs to be = 9hrs, but rejecting it seemed better to avoid corrupted data feeding into the algorithm.

The 32000 was something I found from some experimentation with Vivaldi starting with two sets of coordinates at the origin and trying to get them to drift to the point where adjustment becomes this large negative value. 9 hours also seems like a conservative but large enough upper bound on the previously unbounded rtt value.

### DIFF
--- a/coordinate/client.go
+++ b/coordinate/client.go
@@ -8,8 +8,6 @@ import (
 	"time"
 )
 
-const MAX_RTT_SECONDS = 32400 //9 hours is the upper bound for ping period
-
 // Client manages the estimated network coordinate for a given node, and adjusts
 // it as the node observes round trip times and estimated coordinates from other
 // nodes. The core algorithm is based on Vivaldi, see the documentation for Config
@@ -207,9 +205,9 @@ func (c *Client) Update(node string, other *Coordinate, rtt time.Duration) (*Coo
 		return nil, err
 	}
 
-	durSec := rtt.Seconds()
-	if durSec <= 0 || durSec > MAX_RTT_SECONDS {
-		return nil, fmt.Errorf("Round trip time not in valid range, duration %v is not a positive value less than %v ", durSec, MAX_RTT_SECONDS)
+	const maxRTT = 10 * time.Second
+	if rtt <= 0 || rtt > maxRTT {
+		return nil, fmt.Errorf("round trip time not in valid range, duration %v is not a positive value less than %v ", rtt, maxRTT)
 	}
 
 	rttSeconds := c.latencyFilter(node, rtt.Seconds())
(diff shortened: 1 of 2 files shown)


coordinate/client.go
```diff
@@ -205,6 +207,11 @@ func (c *Client) Update(node string, other *Coordinate, rtt time.Duration) (*Coo
 		return nil, err
 	}
 
+	durSec := rtt.Seconds()
```
slackpad: You can do this as a `time.Duration` (I'd put the constant right here, too):

```
const maxRTT = 10 * time.Second
if rtt < 0 || rtt > maxRTT {
    return nil, fmt.Errorf("RTT %s out of range 0 to %s", rtt, maxRTT)
}
```

slackpad: I'd also rail it way lower since 10 seconds is already several times around the Earth, so this should keep the coordinates better behaved in the face of weirdness (and it's not useful to model anything slower).

coordinate/client_test.go
```diff
@@ -64,6 +65,60 @@ func TestClient_Update(t *testing.T) {
...
+	other := NewCoordinate(config)
+	other.Vec[2] = 0.001
+	rtt := time.Duration(2.0 * other.Vec[2] * secondsToNanoseconds)
+	c, err = client.Update("node", other, rtt)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+
+	dist_old := client.DistanceTo(other)
+	// Update with a valid ping period, should have an affect on estimated rtt
```
slackpad: "effect"

coordinate/client_test.go
```diff
@@ -64,6 +65,60 @@ func TestClient_Update(t *testing.T) {
...
+func TestClient_InvalidInPingValues(t *testing.T) {
+	config := DefaultConfig()
+	config.Dimensionality = 3
+
+	client, err := NewClient(config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Make sure the Euclidean part of our coordinate is what we expect.
```
slackpad: This upper part doesn't seem necessary since we test the coordinate client elsewhere. I'd just try to update with a negative and a too large value and make sure you get the expected error back.

slackpad:
Related to #487.

PR #490, opened 2017-10-20
Removed coordinates from snapshots and ignores any coordinate lines w…

preetapan:
…hen reading snapshot files.


slackpad:
Related to #487 as this provides a recovery mechanism for coordinates in a bad state by simply restarting the agent.

PR #497, opened 2017-12-22
Homebrew no longer uses cask to install serf

grempe:
```
$ brew search serf
==> Searching local taps...
serf
==> Searching taps on GitHub...
==> Searching blacklisted, migrated and deleted formulae...

If you meant "serf" specifically:
It was migrated from caskroom/cask to homebrew/core.
```


slackpad:
Thanks!

Issue #500, opened 2018-01-19
force-leave is not validating input


grounded042:
This is in response to https://github.com/hashicorp/consul/issues/3757. When using the force-leave command it does not give any visual indication to the user if the node is not a member of the cluster. I'm creating this issue so I can submit a PR.

PR #501, opened 2018-01-19
check node cluster membership before force leaving

grounded042:
This is for issue #500.

The core of this is to add a new function CheckAndRemoveFailedNode which checks if a node is a member of the cluster before removing it. I opted to create a new function so as not to break compatability with Consul. Once this gets approval I can update the functionality in Consul.

### DIFF
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -741,6 +741,19 @@ func (s *Serf) Members() []Member {
 	return members
 }
 
+// CheckAndRemoveFailedNode checks if a node exists and then forcibly removes
+// the node via RemoveFailedNode. If the node does not exist in the cluster, it
+// will return exists as false and will not try to remove the node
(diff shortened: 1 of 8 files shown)


serf/serf.go
```diff
@@ -741,6 +741,19 @@ func (s *Serf) Members() []Member {
...
 }
 
+// CheckAndRemoveFailedNode checks if a node exists and then forcibly removes
+// the node via RemoveFailedNode. If the node does not exist in the cluster, it
+// will return exists as false and will not try to remove the node
+func (s *Serf) CheckAndRemoveFailedNode(node string) (exists bool, err error) {
+	// Check to make sure that the node exists in the cluster
+	_, ok := s.members[node]
+	if !ok {
+		return false, nil
```
grounded042: I am not sure if it is ok to return here or if we should check for membership and keep going with the removal even if it's not in the member list. I'm not sure if `s.members` is considered a complete list or if it _might_ not be 100% up to date.

Issue #502, opened 2018-02-12
[docs] Link to SWIM paper is broken


sslavic:
https://www.serf.io/docs/internals/gossip.html has a link to SWIM paper, and it appears to be broken.

PR #503, opened 2018-02-12
Fix broken link to SWIM paper

sslavic:
This fixes https://github.com/hashicorp/serf/issues/502


sslavic:
Took this URL from https://www.consul.io/docs/internals/gossip.html

kyhavlov:
Thanks!

PR #504, opened 2018-02-20
Make ForceLeave and RemoveFailedNode fail when node doesn't exist

### LINKED ISSUE #500: force-leave is not validating input

i0rek:
This PR fixes #500 and https://github.com/hashicorp/consul/issues/3757. I added a test and mentioned the return code in the docs since it changes the CLI return codes.

I notices a couple of things during implementing and testing the changes: 

1) when there are no members, serf doesn't do anything. But after this change I would expect that it says that there are no nodes and that force leaving is not possible.
2) since consul uses `RemoveFailedNode`, consul's CLI is changing as well. Before it returned 0, now it returns 1 if you try to remove a node that doesn't exist.
Before: 
```
MacBook-Pro% consul force-leave abc; echo $?;
0
```
After:
```
MacBook-Pro% consul force-leave abc; echo $?;
Error force leaving: Unexpected response code: 500 (node not found: abc)
1
```

3) I noticed that a node cannot `force-leave` itself

### DIFF
--- a/cmd/serf/command/force_leave_test.go
+++ b/cmd/serf/command/force_leave_test.go
@@ -121,4 +121,8 @@ WAIT:
 	if code != 1 {
 		t.Fatalf("bad: %d. %#v", code, ui.ErrorWriter.String())
 	}
+
+	if !strings.Contains(ui.ErrorWriter.String(), "node not found: invalidNode") {
+		t.Fatalf("bad: %#v", ui.ErrorWriter.String())
+	}
 }
--- a/cmd/serf/command/force_leave_test.go
+++ b/cmd/serf/command/force_leave_test.go
@@ -82,3 +82,43 @@ func TestForceLeaveCommandRun_noAddrs(t *testing.T) {
 		t.Fatalf("bad: %#v", ui.ErrorWriter.String())
 	}
 }
+
+func TestForceLeaveCommandRun_invalidNode(t *testing.T) {
+	a1 := testAgent(t)
+	a2 := testAgent(t)
+	defer a1.Shutdown()
+	defer a2.Shutdown()
+	rpcAddr, ipc := testIPC(t, a1)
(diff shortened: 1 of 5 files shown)


i0rek:
In this PR I changed force-leave to complain about non-existing nodes. That doesn't fix https://github.com/hashicorp/consul/issues/3757 because it is using another method. This is the reason I closed this PR, but maybe it is useful nevertheless?!

i0rek:
I added the check to `RemoveFailedNode` as well.

i0rek:
The build failure should be unrelated. I ran it successfully on my machine.

cmd/serf/command/force_leave_test.go
```diff
@@ -85,3 +86,43 @@ func TestForceLeaveCommandRun_noAddrs(t *testing.T) {
...
+
+	ui := new(cli.MockUi)
+	c := &ForceLeaveCommand{Ui: ui}
+	args := []string{
+		"-rpc-addr=" + rpcAddr,
+		"invalidNode",
+	}
+
+	code := c.Run(args)
+	if code != 1 {
```
kyhavlov: could you check the output of `ui.ErrorWriter.String()` and make sure it has the 'node not found' text?

i0rek: done.

grounded042:
@kyhavlov just curious why this is being reviewed when I've had a PR (#501) in for this for a while now that was never reviewed?

i0rek:
ping @kyhavlov 
@grounded042 maybe it is partially my fault too. I only discovered your work after I finished my PR.

PR #510, opened 2018-05-02
Ensure node leave intents are handled correctly

preetapan:
This PR adds a configurable leave timeout to allow leave intent broadcasts to propagate before returning from the Leave method. 

It also fixes an edge case in  anti entropy sync where a left member in the push/pull message's payload would not get processed because its lamport time was incorrectly set.

Brings in work done in an experimental branch by @slackpad

### DIFF
--- a/serf/serf_test.go
+++ b/serf/serf_test.go
@@ -1345,9 +1345,9 @@ func TestSerf_Leave_SnapshotRecovery(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 	time.Sleep(s2Config.MemberlistConfig.ProbeInterval * 5)
-
-	// Verify that s2 is "left"
-	testMember(t, s1.Members(), s2Config.NodeName, StatusLeft)
+	// Verify that s2 is not in the memberlist
+	// The join intent is ignored because the leave sleeps long enough for the leave intent to broadcast
+	testMember(t, s1.Members(), s2Config.NodeName, StatusNone)
 
 	// Restart s2 from the snapshot now!
 	s2Config.EventCh = nil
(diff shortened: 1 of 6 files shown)


serf/serf_test.go
```diff
@@ -174,6 +174,9 @@ func TestSerf_eventsLeave(t *testing.T) {
 	eventCh := make(chan Event, 4)
 	s1Config := testConfig()
 	s1Config.EventCh = eventCh
+	// Make the reap interval longer in this test
+	// so that the leave does not also cause a reap
+	s1Config.ReapInterval = 30 * time.Second
```
kyhavlov: Was this hiding the issue with the leave not being propagated in this test?

preetapan: Not exactly,  default reap interval in the test is too small, and the leave now waits before the method returns so it was causing this test to fail because the left node also got reaped.

serf/serf_test.go
```diff
@@ -1345,9 +1348,9 @@ func TestSerf_Leave_SnapshotRecovery(t *testing.T) {
 		t.Fatalf("err: %s", err)
 	}
 	time.Sleep(s2Config.MemberlistConfig.ProbeInterval * 5)
-
-	// Verify that s2 is "left"
-	testMember(t, s1.Members(), s2Config.NodeName, StatusLeft)
+	// Verify that s2 is not in the memberlist
+	// The join intent is ignored because the leave sleeps long enough for the leave intent to broadcast
+	testMember(t, s1.Members(), s2Config.NodeName, StatusNone)
```
armon: @preetapan This doesn't seem right to me. After S2 initiates a leave, it shouldn't disappear as a member until it is reaped. Seems like we expect it to be Left state?

preetapan: This was addressed by changing the test setup to increase ReapInterval in #511

PR #511, opened 2018-05-07
Fix test setup for snapshot restore test

preetapan:
This PR fixes a test I changed in PR#510 - instead of asserting that a left node was reaped, the test's reap interval is increased so that we can continue to assert that the node left.


armon:
LGTM!

PR #519, opened 2018-05-28
config: Allow configuration of all memberlist options

jen20:
This pull request is a rebased and modified version of the code submitted by @saromanov in #341. It has been rebased onto the current master branch, and the review comments originally made by @slackpad have been addressed.

### DIFF
--- a/cmd/serf/command/agent/command.go
+++ b/cmd/serf/command/agent/command.go
@@ -316,8 +316,8 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
 			return nil
 		}
 
-		if profileconfig.TCPTimeout != 0 {
-			serfConfig.MemberlistConfig.TCPTimeout = profileconfig.TCPTimeout
+		if profileconfig.StreamTimeout != 0 {
+			serfConfig.MemberlistConfig.TCPTimeout = profileconfig.StreamTimeout
 		}
 		if profileconfig.IndirectChecks != 0 {
 			serfConfig.MemberlistConfig.IndirectChecks = profileconfig.IndirectChecks
--- a/cmd/serf/command/agent/command.go
+++ b/cmd/serf/command/agent/command.go
@@ -302,6 +302,46 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
 		return nil
 	}
 
+	if config.ProfilePath != "" {
+		path := config.ProfilePath
+		f, err := os.Open(path)
+		defer f.Close()
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Error reading '%s': %s", path, err))
+			return nil
+		}
+		profileconfig, err := DecodeProfile(f)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("ProfilePath: %v", err))
+			return nil
+		}
+
+		if profileconfig.TCPTimeout != 0 {
+			serfConfig.MemberlistConfig.TCPTimeout = profileconfig.TCPTimeout
+		}
+		if profileconfig.IndirectChecks != 0 {
+			serfConfig.MemberlistConfig.IndirectChecks = profileconfig.IndirectChecks
+		}
+		if profileconfig.SuspicionMult != 0 {
+			serfConfig.MemberlistConfig.SuspicionMult = profileconfig.SuspicionMult
+		}
+		if profileconfig.PushPullInterval != 0 {
+			serfConfig.MemberlistConfig.PushPullInterval = profileconfig.PushPullInterval
+		}
+		if profileconfig.ProbeTimeout != 0 {
+			serfConfig.MemberlistConfig.ProbeTimeout = profileconfig.ProbeTimeout
+		}
+		if profileconfig.ProbeInterval != 0 {
+			serfConfig.MemberlistConfig.ProbeInterval = profileconfig.ProbeInterval
+		}
+		if profileconfig.GossipNodes != 0 {
+			serfConfig.MemberlistConfig.GossipNodes = profileconfig.GossipNodes
+		}
+		if profileconfig.GossipInterval != 0 {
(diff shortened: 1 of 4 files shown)


cmd/serf/command/agent/config.go
```diff
@@ -127,6 +127,9 @@ type Config struct {
 	// are "wan", "lan", and "local". The default is "lan"
 	Profile string `mapstructure:"profile"`
 
+	// ProfilePath is path to .json file for configuration memberlist
+	ProfilePath string `mapstructure:"profile_path"`
```
kyhavlov: What's the benefit to putting this in a separate json file instead of making it just a section of the config?

website/source/docs/agent/options.html.markdown
```diff
@@ -339,6 +349,98 @@ The remaining keys in the list are considered secondary and are used for
...
+
+The following configuration keys may be used in a configuration file referenced
+using the `profile_path` option in Serf configuration.
+
+Options with a `timeout`, `time` or `interval` suffix may be expressed using
+common unit suffixes for time units - for example `500ms`, `10s` and `1m`.
+  
+* `awareness_max_mult` - Increases the probe interval if the node becomes aware
+  that it might be degraded and not meeting the soft real time requirements to
+  reliably probe other nodes.
```
kyhavlov: Do you think it makes sense to include the defaults for these values here? We do that with the other config docs but it's a little weird here because they come from the memberlist repo.

website/source/docs/agent/options.html.markdown
```diff
@@ -339,6 +349,98 @@ The remaining keys in the list are considered secondary and are used for
...
+  increased bandwidth.
+  
+* `gossip_nodes` - The number of random nodes to send gossip messages to per 
+  `GossipInterval`. Increasing this number causes the gossip messages to propagate
+  across the cluster more quickly at the expense of increased bandwidth.
+
+* `gossip_to_the_dead_time` - the interval after which a node has died that
+  we will still try to gossip to it. This gives the node a chance to refute.
+
+* `indirect_checks` - the number of nodes that will be asked to perform
```
kyhavlov: `- the..` should be capitalized, same for the above

website/source/docs/agent/options.html.markdown
```diff
@@ -339,6 +349,98 @@ The remaining keys in the list are considered secondary and are used for
...
+  
+  ```
+  Retransmits = RetransmitMult * log(N+1)
+  ```
+  
+  This allows the retransmits to scale properly with cluster size. The 
+  higher the multiplier, the more likely a failed broadcast is to converge 
+  at the expense of increased bandwidth.
+
+* `stream_timeout` - is the timeout for establishing a stream connection with a
```
kyhavlov: `- is` should be capitalized

jen20: Should be `s/is t/T/`, I'll fix this.

website/source/docs/agent/options.html.markdown
jen20: :+1:

cmd/serf/command/agent/config.go
jen20: I was wondering about that, but simply migrated the existing work which @slackpad had already reviewed without changing it too much. I'm tempted to suggest that these values should be merged into the main `Config` structure instead such that they can be placed in _any_ JSON file in the configuration directory?

kyhavlov: Yeah that's what I was thinking, replacing ProfilePath with a Profile struct seems easier to use. The only downside would be not being able to configure these things via command line, which is probably fine.

jen20: I think that's a worthwhile change. There is a related issue about doing the same for the key ring elsewhere, though that can be done as a separate step.

website/source/docs/agent/options.html.markdown
jen20: We could include defaults here, but the defaults vary depending on the profile. Given that these are somewhat advanced options, I'm not sure it doesn't make sense to include a link to the relevant part of the memberlist codebase instead?

kyhavlov: That sounds like a good alternative 👍

jen20:
@kyhavlov I've moved the configuration of the memberlist parameters into the main configuration struct, under a `profile_overrides` object, and updated the documentation and tests to reflect this. The documentation nits are also fixed.

cmd/serf/command/agent/command.go
```diff
@@ -302,6 +302,43 @@ func (c *Command) setupAgent(config *Config, logOutput io.Writer) *Agent {
...
+	if config.ProfileOverrides.RetransmitMult != 0 {
+		serfConfig.MemberlistConfig.RetransmitMult = config.ProfileOverrides.RetransmitMult
+	}
+	if config.ProfileOverrides.SuspicionMult != 0 {
+		serfConfig.MemberlistConfig.SuspicionMult = config.ProfileOverrides.SuspicionMult
+	}
+	if config.ProfileOverrides.SuspicionMaxTimeoutMult != 0 {
+		serfConfig.MemberlistConfig.SuspicionMaxTimeoutMult = config.ProfileOverrides.SuspicionMaxTimeoutMult
+	}
+	if config.ProfileOverrides.PushPullInterval != 0 {
```
kyhavlov: In the docs section we say that setting this to 0 will disable it, but that won't actually happen this way - the check should probably be `config.ProfileOverrides.PushPullIntervalRaw != ""` instead (the same goes for GossipInterval below) both here and in MergeConfig.

jen20: Good call, I've made a new commit to update these.

jeromegn:
What happened here?

This sounds useful. We've had a few issues with old events and I thought tweaking these settings would help. Right now I've tweaked the code and built serf msyelf. I might try to pull in this PR.

I assume this died because serf is mainly used in embedded in other programs, which can tweak that configuration easily.

PR #525, opened 2018-08-08
Snapshot compaction threshold is relative to node count

armon:
This PR replaces a static maximum snapshot size with a relative size based on the number of nodes. This is to avoid frequent compaction for large clusters.

Depends on #524

### DIFF
--- a/serf/snapshot.go
+++ b/serf/snapshot.go
@@ -25,10 +25,27 @@ nodes to re-join, as well as restore our clock values to avoid replaying
 old events.
 */
 
-const flushInterval = 500 * time.Millisecond
-const clockUpdateInterval = 500 * time.Millisecond
-const tmpExt = ".compact"
-const snapshotErrorRecoveryInterval = 30 * time.Second
+const (
+	// flushInterval is how often we force a flush of the snapshot file
+	flushInterval = 500 * time.Millisecond
+
+	// clockUpdateInterval is how often we fetch the current lamport time of the cluster and write to the snapshot file
+	clockUpdateInterval = 500 * time.Millisecond
+
+	// tmpExt is the extention we use for the temporary file during compaction
+	tmpExt = ".compact"
+
+	// snapshotErrorRecoveryInterval is how often we attempt to recover from
+	// errors writing to the snapshot file.
+	snapshotErrorRecoveryInterval = 30 * time.Second
+
+	// eventChSize is the size of the event buffers between Serf and the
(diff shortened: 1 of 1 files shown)


serf/snapshot.go
```diff
@@ -25,10 +25,34 @@ nodes to re-join, as well as restore our clock values to avoid replaying
...
+
+	// eventChSize is the size of the event buffers between Serf and the
+	// consuming application. If this is exhausted we will block Serf and Memberlist.
+	eventChSize = 2048
+
+	// shutdownFlushTimeout is the time limit to write pending events to the snapshot during a shutdown
+	shutdownFlushTimeout = 250 * time.Millisecond
+
+	// snapshotBytesPerNode is an estimated bytes per node to snapshot
+	snapshotBytesPerNode = 128
```
banks: This is conservative but that seems like a good thing.

For (my own) reference the format for most entries is:

```
alive: [node name] [ipaddres]:[port]
```

Typically for ipv4 addresses we need 7 + 21 + len(node name). IPv6 is up to 47 bytes. That leaves 74 bytes for hostname in worst case which is plenty.

serf/snapshot.go
```diff
@@ -25,10 +25,34 @@ nodes to re-join, as well as restore our clock values to avoid replaying
...
+
+	// snapshotCompactionThreshold is the threshold we apply to
+	// the snapshot size estimate (nodes * bytes per node) before compacting.
+	snapshotCompactionThreshold = 3
```
banks: This seems kinda high to me? I mean I don't think it's a super big deal but just want to sanity check the edge cases here.

In an example with 1000 nodes, our estimate for the size is already probably ~2x reality for typical hostnames and ipv4 (see last comment). Which means we can probably write 5-6k log entries so it would take a really long time to hit compaction on a somewhat stable cluster.

Erring on the cautious side to minimise disk IO is a good call, but I wonder if for really huge clusters we'll end up with snapshots that actually take significant time to read in on serf startup again?

Back of envelope maths: a cluster with 50k nodes would not compact until it's snapshot is `30000 * 128 * 3 = 19.2MB` which in typical cases (60 byte records) would mean something like 320k serf events recorded. I doubt that's a huge deal to read on startup.

Seems a factor of 2 would be plenty to minimize IO especially given the min size to prevent really small clusters from doing it constantly.

This is a bit of a bikeshed - it's fine like this, just trying to think why it needs to be that high?

armon:
@banks Made the threshold less conservative. Feel free to add a hook test after merge!

PR #529, opened 2018-08-31
Make event compression optional

syndbg:
Adds CLI argument `enable-compression=<bool>` to `serf agent`.

This is to make debugging much easier w/o having LZW compression of messages by `memberlist`.

Signed-off-by: Anton Antonov <anton.synd.antonov@gmail.com>

@banks

### DIFF
--- a/cmd/serf/command/agent/command.go
+++ b/cmd/serf/command/agent/command.go
@@ -55,6 +55,8 @@ func (c *Command) readConfig() *Config {
 	var tags []string
 	var retryInterval string
 	var broadcastTimeout string
+	var disableCompression bool
+
 	cmdFlags := flag.NewFlagSet("agent", flag.ContinueOnError)
 	cmdFlags.Usage = func() { c.Ui.Output(c.Help()) }
 	cmdFlags.StringVar(&cmdConfig.BindAddr, "bind", "", "address to bind listeners to")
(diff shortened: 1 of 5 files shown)


syndbg:
@banks 

```
In other words can we change the option to be -disable-compression which defaults to false. It seems much more obvious then that compression is the default and the singular flag -disable-compression without =<bool> can be used to disable it which is standard UX for our tools.
```

Ok, will do.

```
I realise you took the name from the Memberlist config which does the exact opposite of what I just said but I still think this is a better/more consistent UX for Serf vs a library struct.
```

Yep, but I don't mind. I get your point with the UX of just passing `-disable-compression`.


Will change it in a bit after few meetings.

syndbg:
@banks Updated & ready for review.

Added the points you made, also added UI message whether Message Compression is enabled or not.

cmd/serf/command/agent/command.go
```diff
@@ -92,11 +92,22 @@ func (c *Command) readConfig() *Config {
 	cmdFlags.StringVar(&retryInterval, "retry-interval", "", "retry join interval")
 	cmdFlags.BoolVar(&cmdConfig.RejoinAfterLeave, "rejoin", false,
 		"enable re-joining after a previous leave")
+
+	var disableCompression bool
```
banks: _super_ nitpicky but all the other free variables used for reading options are defined up at the start of this method. I don't have a strong preference either way and can see benefits to doing it here but I feel like being consistent with existing code is probably more important than any other potential benefits!

cmd/serf/command/agent/command.go
```diff
@@ -452,6 +463,7 @@ func (c *Command) startAgent(config *Config, agent *Agent,
 	c.Ui.Info(fmt.Sprintf("     Encrypted: %#v", agent.serf.EncryptionEnabled()))
 	c.Ui.Info(fmt.Sprintf("      Snapshot: %v", config.SnapshotPath != ""))
 	c.Ui.Info(fmt.Sprintf("       Profile: %s", config.Profile))
+	c.Ui.Info(fmt.Sprintf("       Message Compression Enabled: %v", config.EnableCompression))
```
banks: Another nit: this breaks the right-aligned formatting which is useful for quick skimming as well as just looking odd. I'd suggest we shoot for the following output instead:

```
           Encrypted: true
            Snapshot: /foo/bar
             Profile: ...
 Messages Compressed: true
```

That is with an extra six spaces prefix the original items and a single space prefix for the new one to differentiate for other log lines below.

website/source/docs/agent/options.html.markdown
```diff
@@ -141,6 +141,8 @@ The options below are all specified on the command-line.
 * `-retry-max` - Provides a limit on how many attempts to join the cluster
   can be made by `-retry-join`. If 0, there is no limit, and the agent will
   retry forever. Defaults to 0.
+  
+* `-disable-compression` - Disable message compression for broadcasting events. Enabled by default.
```
banks: Maybe add "This is generally only useful for debugging."

PR #530, opened 2018-09-05
 Make `UserEventSizeLimit` configurable and fix `UserEventSizeLimit` check

syndbg:
Different usages, needs different `UserEventSizeLimit`.
It's understood that having a big event size is going to make the
broadcast and replication across the cluster slow depending on the size.

Other serf broadcast timeout and replication timeout settings must be configured accordingly!

Also fix the check:

* Make it more reliable in terms of payload size since only the encoded message is checked.
* Make it consistent with `QuerySizeLimit`.

@banks Another one. Although I require some HashiCorp knowledge why it wasn't configurable before. My idea is to piggyback on the `UserEvent` to send application events that may be greater than `512` bytes, but in any case they must be configurable. `QuerySizeLimit` has been taken into account as another example how it is risky, yet configurable.

P.S https://media.giphy.com/media/GV3aYiEP8qbao/giphy.gif

### DIFF
--- a/cmd/serf/command/agent/config.go
+++ b/cmd/serf/command/agent/config.go
@@ -35,6 +35,7 @@ func DefaultConfig() *Config {
 		SyslogFacility:         "LOCAL0",
 		QueryResponseSizeLimit: 1024,
 		QuerySizeLimit:         1024,
+		UserEventSizeLimit:     512,
 		BroadcastTimeout:       5 * time.Second,
 	}
 }
@@ -113,6 +114,10 @@ type Config struct {
 	// configuration.
 	QueryResponseSizeLimit int `mapstructure:"query_response_size_limit"`
 	QuerySizeLimit         int `mapstructure:"query_size_limit"`
+
+	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
+	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
+	UserEventSizeLimit int `mapstructure:"user_event_size_limit"`
 
 	// StartJoin is a list of addresses to attempt to join when the
 	// agent starts. If Serf is unable to communicate with any of these
@@ -462,6 +467,9 @@ func MergeConfig(a, b *Config) *Config {
 	}
 	if b.QuerySizeLimit != 0 {
 		result.QuerySizeLimit = b.QuerySizeLimit
+	}
+	if b.UserEventSizeLimit != 0 {
+		result.UserEventSizeLimit = b.UserEventSizeLimit
 	}
 	if b.BroadcastTimeout != 0 {
 		result.BroadcastTimeout = b.BroadcastTimeout
--- a/serf/config.go
+++ b/serf/config.go
@@ -242,6 +242,10 @@ type Config struct {
 	// Merge can be optionally provided to intercept a cluster merge
 	// and conditionally abort the merge.
 	Merge MergeDelegate
+
+	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
+	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
+	UserEventSizeLimit int
 }
 
 // Init allocates the subdata structures
@@ -282,5 +286,6 @@ func DefaultConfig() *Config {
 		QuerySizeLimit:               1024,
 		EnableNameConflictResolution: true,
 		DisableCoordinates:           false,
+		UserEventSizeLimit:           512,
 	}
 }
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -223,8 +223,7 @@ type queries struct {
 }
 
 const (
-	UserEventSizeLimit = 512        // Maximum byte size for event name and payload
-	snapshotSizeLimit  = 128 * 1024 // Maximum 128 KB snapshot
+	snapshotSizeLimit = 128 * 1024 // Maximum 128 KB snapshot
 )
 
 // Create creates a new Serf instance, starting all the background tasks
@@ -437,14 +436,13 @@ func (s *Serf) KeyManager() *KeyManager {
 }
 
 // UserEvent is used to broadcast a custom user event with a given
-// name and payload. The events must be fairly small, and if the
-// size limit is exceeded and error will be returned. If coalesce is enabled,
-// nodes are allowed to coalesce this event. Coalescing is only available
-// starting in v0.2
+// name and payload. If the configured size limit is exceeded and error will be returned.
+// If coalesce is enabled, nodes are allowed to coalesce this event.
+// Coalescing is only available starting in v0.2
 func (s *Serf) UserEvent(name string, payload []byte, coalesce bool) error {
 	// Check the size limit
-	if len(name)+len(payload) > UserEventSizeLimit {
-		return fmt.Errorf("user event exceeds limit of %d bytes", UserEventSizeLimit)
+	if len(name)+len(payload) > s.config.UserEventSizeLimit {
+		return fmt.Errorf("user event exceeds limit of %d bytes", s.config.UserEventSizeLimit)
 	}
 
 	// Create a message
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -224,6 +224,7 @@ type queries struct {
 
 const (
 	snapshotSizeLimit = 128 * 1024 // Maximum 128 KB snapshot
+	UserEventSizeLimit = 9 * 1024  // Maximum 9KB for event name and payload
 )
 
 // Create creates a new Serf instance, starting all the background tasks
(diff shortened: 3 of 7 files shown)


serf/serf.go
```diff
@@ -437,33 +436,34 @@ func (s *Serf) KeyManager() *KeyManager {
...
 	// Start broadcasting the event
 	raw, err := encodeMessage(messageUserEventType, &msg)
 	if err != nil {
 		return err
 	}
+
+	// Check the size
+	if len(raw) > s.config.UserEventSizeLimit {
+		return fmt.Errorf("user event exceeds limit of %d bytes", s.config.UserEventSizeLimit)
+	}
```
banks: I don't think we should move this check - if user maliciously or due to a bug passes a 1GB payload then doing all the work to allocate and encode that before returning the error is wasted.

In practice the encoding overhead is small and fixed-size so anything more than a few bytes is almost entirely dominated by the user payload (and name) so failing earlier seems better to me.

syndbg:
@banks 

```
I'd strongly advise we have a sane upper limit that stops users incorrectly configuring the event size to be more than will fit in a UDP packet. Technically that varies depending on path MTU which we can't know (another great reason to not let this be configurable at all) 
```

Well, the same flaw can be said for queries. It can be observed at https://github.com/hashicorp/serf/blob/master/serf/serf.go#L526 .

To be honest, queries and user events must be TCP in the first place. Even the queries ack is still going to mimick TCP a bit.

In the longer run, I'm doing this step just to detach Serf a bit from UDP-coupling and to enable me to use TCP only if I wish to and/or switch between TCP or UDP on demand.

It boils down to 3 strategies that we've discussed for our usages of `serf` as a base:

1. Use UDP only and eventually TCP resync - current behavior, no changes whatsoever required.
2. Use TCP for everything that is currently UDP and keep the eventual TCP resync as it is.
3. Use UDP only for membership gossip, use TCP for queries and user events and keep the TCP resync as it is.

`I'd say a hard upper limit of 9kb is sensible for now.`

What do you think about applying it for both queries and user events? 

```
Please update the documentation in the website dir to note the config option changes (I forgot to ask this in your last PR too so if you could add that here or in a new PR that would be great). I'd prefer if we make it very clear that there are lots of tradeoffs and considerations to take into account like MTU, cluster gossip bandwidth, timeouts and more that may need to be tuned if this is changed and that it's very strongly recommended to keep it conservative.
```

Whoops, missed the website. I'll check where it must be updated to reflect the changes here and in the previous PR.

```
Sending more than a few hundred bytes via Gossip is likely the wrong solution already - it tends to be better to send small messages and use reliable transport to fetch details when the receiver wishes to act on it.
```

Perhaps, perhaps not. Depends on what the goal of the gossip is, in my opinion.

Just out of curiosity, how is this realized in Consul and Nomad? From what I've seen just by poking a bit in the RPC "layer", I see that the pattern there is to send reliable messages via TCP and not to gossip them. I didn't see any RPC gossip at all. 

On the other hand, in the `cmd/serf`, which I'll refer to as example `serf` implementation, there's an agent IPC layer that is again TCP-based, but after receiving the TCP user event, it's still handled via the gossip UDP. Ref: https://github.com/hashicorp/serf/blob/master/cmd/serf/command/agent/agent.go#L186

P.S Do you have some sort of IM communication channel like slack, gitter or whichever? I think this will be faster for both of us to clear our intents and concerns.

banks:
Thanks for the context. We tend not to use IM for public facing issues as we've found it too hard to have engineers balance interruptions from IM if they are always available vs never actually being present there or getting behind on massive backlogs.

The big problem with Gossip to disseminate large messages is that the are inherently amplified. Gossip is super efficient for small messages as it allows huge clusters to efficiently converge, but the cost of that is that each message gets sent many times to multiple nodes to increase the rate of convergence.

So Gossip is well suited for very small "events" like "node 10.10.10.10 joined the cluster" which are a few bytes. But it's a poor fit for sending large payloads like "here is a whole new multi-KB config to store and use".

The recommended approach is to broadcast very small events that give enough info that _something_ happened in a scalable way, but to have any significant payload delivered over some other non-gossip (probably TCP) protocol such as HTTP/RPC/Whatever you app uses already. E.g. "Config was updated to version 10112 on node XXXX" and then the receiver if they care about that can go and request that new config version.

I think that is why we hard coded this as small in the first place to just side step all the potential bad usage and documenting the tradeoffs. I'm not against that provided the documentation is very clear that it's rare that you should need to increase it, and you'll need to spend non-trivial effort testing in your environment whether it actually achieves the goals you have without consuming too much CPU, causing head-of-line blocking that affects cluster stability, misses timeouts too often etc.

> Just out of curiosity, how is this realized in Consul and Nomad?

Serf/Gossip is pretty much only used for membership and failure detection in Consul and Nomad, almost all other communication is done through RPC which is multiplexed over TCP (similar to http2/gRPC but not ). Consul does expose Serf's user events but note the docs here: https://www.consul.io/docs/commands/event.html

> The underlying gossip also sets limits on the size of a user event message. It is hard to give an exact number, as it depends on various parameters of the event, but the payload should be kept very small (< 100 bytes). Specifying too large of an event will return an error.

syndbg:
@banks Sure.

vtolstov:
any chance to get this merged?

serf/serf.go
syndbg: Failing earlier :+1:

syndbg:
@vtolstov Sorry for the lack of progress here.

Actually what was left:

```
my request was just to limit to some sanish upper bound, I still think 9k is more than enough given all the other things mentioned here! Does that seem reasonable?
```

I've just added this in two steps:

1. When creating a `Serf` instance it validate that the specified config is lower than the sane 9kb limit. https://github.com/hashicorp/serf/pull/530/files#diff-d6423c362af3a8e179ae6bf40ff94196R245
1. When sending a message via the `Serf` instance's `UserEvent` function w/ receiver it checks if the payload before encoding exceeds the sanish 9kb limit. https://github.com/hashicorp/serf/pull/530/files#diff-d6423c362af3a8e179ae6bf40ff94196R458
1. When sending a message via the `Serf` instance's `UserEvent` function w/ receiver it checks if the payload after encoding exceeds the sanish 9kb limit. https://github.com/hashicorp/serf/pull/530/files#diff-d6423c362af3a8e179ae6bf40ff94196R488


@banks I think it's ready for review and I can dedicate the time needed to make any code review changes/suggestions.

However, the test suite is failing in `master`. From my end it looks like nothing is broken by my changes.

cmd/serf/command/agent/config.go
```diff
@@ -114,6 +115,10 @@ type Config struct {
 	QueryResponseSizeLimit int `mapstructure:"query_response_size_limit"`
 	QuerySizeLimit         int `mapstructure:"query_size_limit"`
 
+	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
+	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
+	UserEventSizeLimit int `mapstructure:"user_event_size_limit"`
```
banks: I feel like we should document that it is the user's responsibility to ensure this is still lower than their network's minimum path MTU between any two nodes in the cluster otherwise things aren't going to work well (worst case performance decrease but I suspect we won't handle fragmentation well and cause more "failures"). I've not actually validated this is true but it is transmitted by UDP right?

serf/config.go
```diff
@@ -242,6 +242,10 @@ type Config struct {
 	// Merge can be optionally provided to intercept a cluster merge
 	// and conditionally abort the merge.
 	Merge MergeDelegate
+
+	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
+	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
```
banks: Same as above.

cmd/serf/command/agent/config.go
banks: Docs say 

> Because user events are sent along the gossip layer, which uses UDP, the payload and entire message framing must fit within a single UDP packet. 

We could maybe repeat that here and point out that users should only increase that if they know their path MTU will be sufficient.

syndbg:
```
I've not actually validated this is true but it is transmitted by UDP right?
```

Correct, it's UDP :+1: 

```
We could maybe repeat that here and point out that users should only increase that if they know their path MTU will be sufficient.
```

Yep :+1:

syndbg: Updated in 151d9a86e24c8a380f2c9bce17a00d9abd2a9a69

serf/config.go
syndbg: Updated in 151d9a86e24c8a380f2c9bce17a00d9abd2a9a69

syndbg:
ping @banks

PR #545, opened 2019-01-03
ensure performance improvements in memberlist are activated

rboyer:
This is the serf side of https://github.com/hashicorp/memberlist/pull/167/


rboyer:
Note that this is only treating serf as a library. Nothing was done to update the vendored code yet for serf-as-a-binary.

armon:
LGTM

PR #546, opened 2019-01-16
Truncate key list response

i0rek:
This truncates the list key response in case it contains too many keys so that the message would be bigger than allowed and serf wouldn't been able to send it out.

This is what it looks like if you run into it: 

```
2019/01/15 12:51:33 [INFO] serf: Received list-keys query
2019/01/15 12:51:33 [ERR] serf: Failed to respond to key query: response exceeds limit of 1024 bytes
```

If there are too many keys in a response, the keys will be removed from the end until it fits. That way the primary key (index 0) is always part of the response. In case of truncation there will be a warning, stating that it has been truncated. 
The receiving agent checks if the message is close to the limit and emit a warning about truncation as well.

I have to admit that this approach feels a little strange, but I think a truncated response is better than no response at all.

### DIFF
--- a/serf/internal_query.go
+++ b/serf/internal_query.go
@@ -2,6 +2,7 @@ package serf
 
 import (
 	"encoding/base64"
+	"fmt"
 	"log"
 	"strings"
 )
@@ -149,17 +150,60 @@ func (s *serfQueries) handleConflict(q *Query) {
 	}
 }
 
+func (s *serfQueries) keyListResponseWithCorrectSize(q *Query, resp *nodeKeyResponse) ([]byte, messageQueryResponse, error) {
+	warned := false
+	for i := len(resp.Keys); i >= 0; i-- {
+		buf, err := encodeMessage(messageKeyResponseType, resp)
+		if err != nil {
+			return nil, messageQueryResponse{}, err
+		}
+
+		// Create response
+		qresp := q.createResponse(buf)
+
+		// Encode response
+		raw, err := encodeMessage(messageQueryResponseType, qresp)
+		if err != nil {
+			return nil, messageQueryResponse{}, err
+		}
+
+		// Check the size limit
+		if err = q.checkResponseSize(raw); err != nil {
+			resp.Keys = resp.Keys[0:i]
+			if !warned {
+				s.logger.Printf("[WARN] serf: truncated key list list response so that it fits into message")
+				warned = true
+			}
+			continue
+		}
+		return raw, qresp, nil
+	}
+	return nil, messageQueryResponse{}, fmt.Errorf("Failed to truncate response so that it fits response size.")
+}
+
 // sendKeyResponse handles responding to key-related queries.
 func (s *serfQueries) sendKeyResponse(q *Query, resp *nodeKeyResponse) {
-	buf, err := encodeMessage(messageKeyResponseType, resp)
-	if err != nil {
-		s.logger.Printf("[ERR] serf: Failed to encode key response: %v", err)
-		return
-	}
-
-	if err := q.Respond(buf); err != nil {
-		s.logger.Printf("[ERR] serf: Failed to respond to key query: %v", err)
-		return
+	switch q.Name {
+	case internalQueryName(listKeysQuery):
+		raw, qresp, err := s.keyListResponseWithCorrectSize(q, resp)
+		if err != nil {
+			s.logger.Printf("[ERR] serf: %v", err)
+			return
+		}
+		if err := q.respondWithMessageAndResponse(raw, qresp); err != nil {
+			s.logger.Printf("[ERR] serf: Failed to respond to key query: %v", err)
+			return
+		}
+	default:
+		buf, err := encodeMessage(messageKeyResponseType, resp)
+		if err != nil {
+			s.logger.Printf("[ERR] serf: Failed to encode key response: %v", err)
+			return
+		}
+		if err := q.Respond(buf); err != nil {
+			s.logger.Printf("[ERR] serf: Failed to respond to key query: %v", err)
+			return
+		}
 	}
 }
 
--- a/serf/internal_query.go
+++ b/serf/internal_query.go
@@ -31,7 +31,7 @@ const (
 	listKeysQuery = "list-keys"
 
 	// maxListKeyFactor is used to compute the max number of keys in a list key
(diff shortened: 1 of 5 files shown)


serf/internal_query.go
```diff
@@ -149,17 +150,60 @@ func (s *serfQueries) handleConflict(q *Query) {
...
+		// Check the size limit
+		if err = q.checkResponseSize(raw); err != nil {
+			resp.Keys = resp.Keys[0:i]
+			if !warned {
+				s.logger.Printf("[WARN] serf: truncated key list list response so that it fits into message")
+				warned = true
+			}
+			continue
+		}
+		return raw, qresp, nil
```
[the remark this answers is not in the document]
banks: So it looks like the payload is: https://github.com/hashicorp/serf/blob/6a7f035e9efd9b99d33ee452f0b26501faec7d52/serf/internal_query.go#L51-L60

Which I notice two things:

 1. It has a Message field - could we return the fact that the truncation occured there? It would allow us to be sure in the client rather than guessing if it's close. (It adds bytes though of course.
 2. It's pretty simple although it is msgPack encoded which is not totally trivial (varints for example). In this case everything is a string or fixed length though and we could rely on all keys being the same length too so we can probably make a hand-wavy heuristic for how many keys will fit:
  - 1 byte for the struct/map header (since it's only 3 elements)
  - 1 byte for (short) string key header + 6 bytes for "result"
  - 1 byte for the bool
  - 1 byte for (short) string key header + 7 bytes for "message"
  - 1 byte for the empty string header (plus more if we use this for error per 1 above)
  - 1 byte for the (short) string key header + 4 bytes for "keys"
  - 5 bytes for the list header (3 bytes is enough up to 65k keys but be conservative)
  - so far I make that 28 bytes with empty Message
  - N * 1 byte for string header (keys are shorter than 31 bytes in base64) + 24 byte for base64 key with padding

You could add some fudge factors there but my estimate is that if the max query size is default 1024 then you can get 39 keys in.

I think in practice people were seeing issues with a few less - it's not clear if that's down to me missing something above (maybe we already use the Message for something).

I don't know if it makes sense to try to to exactly calculate message size, but my point is we know it's about 30 something keys so if there are 10,000 keys present then iterating and re-encoding 9,960+ times is not super clever!

We could maybe have a really simple model that uses big fudge factors (like `QueryResponseSizeLimit/24` where 24 is minimum number of bytes for a 16 byte key in base64) as the starting point (because we know for sure we can't encode more than that. Then we'd only have to iterate at worst a handfull of times rather than unbounded?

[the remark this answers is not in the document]
i0rek: Thanks for suggestion! I wrote a test that estimates that factor: TestSerfQueries_estimateMaxKeysInListKeyResponseFactor. I took an unrealistic message size of 10240 to make sure we err on the side of being to generous and also to account for other configurations.

In case of the default size, the algorithm starts at index 40, no matter how many keys there are.

serf/keymanager.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sync"
 )
 
+const MagicTruncationNumber = 50
```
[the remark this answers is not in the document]
i0rek: Made obsolete by using the message field.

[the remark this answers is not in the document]
i0rek: I hesitated using the message field because so far it was only used for sending errors, but I figured I could use it for information as well.

serf/internal_query.go
```diff
@@ -29,6 +29,13 @@ const (
 
 	// listKeysQuery is used to list all known keys in the cluster
 	listKeysQuery = "list-keys"
+
+	// maxListKeyFactor is used to compute the max number of keys in a list key
+	// response. eg 1024/25 = 40. a message with max size of 1024 bytes cannot
+	// contain more than 40 keys. There is a test
+	// (TestSerfQueries_estimateMaxKeysInListKeyResponse) which does the
+	// computation and in case of changes, the value can be adjusted.
+	maxListKeyFactor = 25
```
banks: nit: is it worth mentioning that 25 is derived from the length of an AES 16 byte key in base64? I might even call it `minEncodedKeyLength` or something but up to you. Keys are not always 16 bytes necessarily - memberlist supports 32 byte keys to _in theory_ in practice we probably only use 16 bytes but that is the minimum which is what we care about for our estimate.

serf/internal_query.go
```diff
@@ -152,7 +159,8 @@ func (s *serfQueries) handleConflict(q *Query) {
 
 func (s *serfQueries) keyListResponseWithCorrectSize(q *Query, resp *nodeKeyResponse) ([]byte, messageQueryResponse, error) {
 	warned := false
-	for i := len(resp.Keys); i >= 0; i-- {
+	maxListKeys := q.serf.config.QueryResponseSizeLimit / maxListKeyFactor
```
banks: this is a crazy edge case but if `QueryResponseSizeLimit` was set to something less that `maxListKeyFactor` then `maxListKeys` would end up being `0` and the loop below would never run and we'd return...

Oh OK it returns an error at the bottom so it's fine I thought it was going to be an empty message. Fine!

serf/keymanager.go
```diff
@@ -68,6 +68,10 @@ func (k *KeyManager) streamKeyResp(resp *KeyResponse, ch <-chan NodeResponse) {
 			resp.NumErr++
```
banks: Looks like this code will implicitly count the message as implying an error occurred. I'm not sure if that's good or bad - it's an error in the sense that the list produced is incomplete so you can't rely on the output. What's the end-user UX look like? I'm OK if it says its an error and explains but if it for example hides the actual results or say something too dramatic maybe we should reconsider?

i0rek: It is only counted as error if result is false. Since I still return true, even when truncating, it should be fine.
(This was my initial thinking as well and I did not want to make that an error).

banks: Oh cool, in that case does our message actually get seen by the user in the CLI output or just logs? Would be nice if it was in CLI output!

i0rek: Well actually, I didn't think of showing it in the UI! But now it even shows up there with https://github.com/hashicorp/consul/pull/5236.

banks:
Couple of minor nits still unresolved but still good to go as is if you don't think they are important!

PR #548, opened 2019-01-23
Gomodify

jefferai:
Each commit in this series is a single step starting with the current govendor, with the commit message explaining the specific change.


i0rek:
Good point @rboyer! That would make it easier for ppl(like me) who are still running everything in their gopath.

jefferai:
@rboyer @mkeeler @i0rek I've added three more commits with descriptions of each change based on feedback. PTAL!

.travis.yml
```diff
@@ -1,16 +1,18 @@
 language: go
 
 go:
-  - 1.10.x
-  - 1.11.x
+  - 1.11.4
```
rboyer:
```suggestion
  - "1.11.4"
```
I think the quotes are necessary now to avoid yaml weirdness.

GNUmakefile
```diff
@@ -46,12 +46,12 @@ tools::
 
 # updatedeps installs all the dependencies needed to test, build, and run
 updatedeps:: tools
-	govendor list -no-status +vendor | xargs -n1 go get -u
-	govendor update +vendor
+	go get -u
```
rboyer: Should this also have `./...`?

jefferai: Nope!

.travis.yml
jefferai: Ah really? Huh.

rboyer:
I see you nuked `vendor`. I can understand that for libraries in a go-mod world. Was the overall plan to retain the `vendor` directory in deployables for the short term until the caching proxies are ready?

jefferai:
@rboyer I'll put it back for now, tied to gomod in the `updatedeps` target.

PR #563, opened 2019-05-17
fix If raw does not hold a MemberEvent, the statement will trigger  a panic.

withlin:
as title


withlin:
@alvin-huang @mkeeler  ptal, tks.

PR #574, opened 2019-08-29
Trigger Removal for Left/Failed Agents

s-christoff:
Adds the `-prune` flag to the `forceleave` command to allow for an agent to be completely removed from Serf's list of members. This can only be ran with the force leave command, so this can only be ran on nodes in a failed or left state.

### DIFF
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -199,7 +199,8 @@ func (c *RPCClient) ForceLeave(node string) error {
 	return c.genericRPC(&header, &req, nil)
 }
 
-// If the prune flag is set, remove the node entirely
+//ForceLeavePrune uses ForceLeave but is used to reap the
+//node entirely
 func (c *RPCClient) ForceLeavePrune(node string) error {
 	header := requestHeader{
 		Command: forceLeaveCommand,
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -187,14 +187,27 @@ func (c *RPCClient) Close() error {
 
 // ForceLeave is used to ask the agent to issue a leave command for
 // a given node
-func (c *RPCClient) ForceLeave(node string, prune bool) error {
+func (c *RPCClient) ForceLeave(node string) error {
 	header := requestHeader{
 		Command: forceLeaveCommand,
 		Seq:     c.getSeq(),
 	}
 	req := forceLeaveRequest{
 		Node:  node,
-		Prune: prune,
+		Prune: false,
+	}
+	return c.genericRPC(&header, &req, nil)
+}
+
+// If the prune flag is set, remove the node entirely
+func (c *RPCClient) ForceLeavePrune(node string) error {
+	header := requestHeader{
+		Command: forceLeaveCommand,
+		Seq:     c.getSeq(),
+	}
+	req := forceLeaveRequest{
+		Node:  node,
+		Prune: true,
 	}
 	return c.genericRPC(&header, &req, nil)
 }
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -187,13 +187,14 @@ func (c *RPCClient) Close() error {
 
 // ForceLeave is used to ask the agent to issue a leave command for
 // a given node
-func (c *RPCClient) ForceLeave(node string) error {
+func (c *RPCClient) ForceLeave(node string, prune bool) error {
 	header := requestHeader{
 		Command: forceLeaveCommand,
 		Seq:     c.getSeq(),
 	}
 	req := forceLeaveRequest{
-		Node: node,
+		Node:  node,
+		Prune: prune,
 	}
 	return c.genericRPC(&header, &req, nil)
 }
--- a/cmd/serf/command/agent/agent.go
+++ b/cmd/serf/command/agent/agent.go
@@ -182,7 +182,7 @@ func (a *Agent) ForceLeave(node string) error {
 	return err
 }
 
-//ForceLeavePrune completely removes a failed node from the
+// ForceLeavePrune completely removes a failed node from the
 //member list entirely
 func (a *Agent) ForceLeavePrune(node string) error {
 	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
--- a/cmd/serf/command/agent/agent.go
+++ b/cmd/serf/command/agent/agent.go
@@ -185,10 +185,10 @@ func (a *Agent) ForceLeave(node string) error {
 // ForceLeavePrune completely removes a failed node from the
 // member list entirely
 func (a *Agent) ForceLeavePrune(node string) error {
-	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
+	a.logger.Printf("[INFO] agent: Force leaving node (prune): %s", node)
 	err := a.serf.RemoveFailedNodePrune(node)
 	if err != nil {
-		a.logger.Printf("[WARN] agent: failed to remove node: %v", err)
+		a.logger.Printf("[WARN] agent: failed to remove node (prune): %v", err)
 	}
 	return err
 }
--- a/cmd/serf/command/agent/agent.go
+++ b/cmd/serf/command/agent/agent.go
@@ -183,7 +183,7 @@ func (a *Agent) ForceLeave(node string) error {
 }
 
 // ForceLeavePrune completely removes a failed node from the
-//member list entirely
+// member list entirely
 func (a *Agent) ForceLeavePrune(node string) error {
 	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
 	err := a.serf.RemoveFailedNodePrune(node)
--- a/cmd/serf/command/agent/agent.go
+++ b/cmd/serf/command/agent/agent.go
@@ -173,9 +173,9 @@ func (a *Agent) Join(addrs []string, replay bool) (n int, err error) {
 }
 
 // ForceLeave is used to eject a failed node from the cluster
-func (a *Agent) ForceLeave(node string) error {
+func (a *Agent) ForceLeave(node string, prune bool) error {
 	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
-	err := a.serf.RemoveFailedNode(node)
+	err := a.serf.RemoveFailedNode(node, prune)
 	if err != nil {
 		a.logger.Printf("[WARN] agent: failed to remove node: %v", err)
 	}
--- a/cmd/serf/command/agent/agent.go
+++ b/cmd/serf/command/agent/agent.go
@@ -173,9 +173,20 @@ func (a *Agent) Join(addrs []string, replay bool) (n int, err error) {
 }
 
 // ForceLeave is used to eject a failed node from the cluster
-func (a *Agent) ForceLeave(node string, prune bool) error {
+func (a *Agent) ForceLeave(node string) error {
 	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
-	err := a.serf.RemoveFailedNode(node, prune)
+	err := a.serf.RemoveFailedNode(node)
+	if err != nil {
+		a.logger.Printf("[WARN] agent: failed to remove node: %v", err)
+	}
+	return err
+}
+
+//ForceLeavePrune completely removes a failed node from the
+//member list entirely
+func (a *Agent) ForceLeavePrune(node string) error {
+	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
+	err := a.serf.RemoveFailedNodePrune(node)
 	if err != nil {
 		a.logger.Printf("[WARN] agent: failed to remove node: %v", err)
 	}
--- a/cmd/serf/command/agent/ipc.go
+++ b/cmd/serf/command/agent/ipc.go
@@ -124,7 +124,8 @@ type eventRequest struct {
 }
 
 type forceLeaveRequest struct {
-	Node string
+	Node  string
+	Prune bool
 }
 
 type joinRequest struct {
@@ -605,7 +606,7 @@ func (i *AgentIPC) handleForceLeave(client *IPCClient, seq uint64) error {
 	}
 
 	// Attempt leave
-	err := i.agent.ForceLeave(req.Node)
+	err := i.agent.ForceLeave(req.Node, req.Prune)
 
 	// Respond
 	resp := responseHeader{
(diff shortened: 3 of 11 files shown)


serf/serf.go
```diff
@@ -1438,6 +1453,33 @@ func (s *Serf) resolveNodeConflict() {
 	}
 }
 
+//handlePrune is a simplified version of Reap() which only runs when the prune
```
banks: nit (space after slashes):
```suggestion
// handlePrune is a simplified version of Reap() which only runs when the prune
```

serf/serf.go
```diff
@@ -1438,6 +1453,33 @@ func (s *Serf) resolveNodeConflict() {
...
+	delete(s.members, m.Name)
+
+	// Tell the coordinate client the node has gone away and delete
+	// its cached coordinates.
+	if !s.config.DisableCoordinates {
+		s.coordClient.ForgetNode(m.Name)
+
+		s.coordCacheLock.Lock()
+		delete(s.coordCache, m.Name)
+		s.coordCacheLock.Unlock()
```
banks: I don't know this code well so I'll go look but is it possible this could race since we `ForgetNode` above and then take lock on cache? It might be OK in that you should only be using this on nodes you know are really dead but I wonder if it's possible for an out-of-order `alive` message from the node to be delivered (or node to come back up) and recreate the client in `coordClient`, updates it's cached coords but then have them removed again here 🤔.

I _guess_ (again not checked the code) the worst case is that we'd not have coordinates until next message from that node so it's not a huge deal probably but just checking. If this is the same pattern we do elsewhere already (e.g. during leave handling) then cool!

serf/serf.go
```diff
@@ -1438,6 +1453,33 @@ func (s *Serf) resolveNodeConflict() {
...
+	}
+
+	// Send an event along
+	s.logger.Printf("[INFO] serf: EventMemberReap: %s", m.Name)
```
banks: Is it worth making this log different to indicate it was a force reap not a time-based one? Maybe something like:
```suggestion
	s.logger.Printf("[INFO] serf: EventMemberReap (forced): %s", m.Name)
```

serf/serf_test.go
```diff
@@ -836,6 +836,68 @@ func TestSerfRemoveFailedNode(t *testing.T) {
...
+	// Verify that s2 is "failed"
+	testMember(t, s1.Members(), s2Config.NodeName, StatusFailed)
+
+	// Now remove the failed node
+	if err := s1.RemoveFailedNode(s2Config.NodeName, true); err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	// Check to make sure it's gone
+	if len(s1.Members()) != 2 {
```
banks: This seems like it would end up being super flaky given the timing involved. Is it worth using a retry helper around this and the assertion below to allow time for the force reap to propagate?

serf/serf.go
s-christoff: I totally dig this. I was thinking if wondering if I _should_ do something like that and was hoping that someone would comment on it!

client/rpc_client.go
```diff
@@ -187,13 +187,14 @@ func (c *RPCClient) Close() error {
 
 // ForceLeave is used to ask the agent to issue a leave command for
 // a given node
-func (c *RPCClient) ForceLeave(node string) error {
+func (c *RPCClient) ForceLeave(node string, prune bool) error {
```
rboyer: Do we care about breaking callers of this function? If it's just consul that we expect to be using it then it's fine, otherwise you may need to introduce a new function with 2 args and delegate the old function to it.

cmd/serf/command/force_leave.go
```diff
@@ -68,6 +71,7 @@ Options:
 
   -rpc-addr=127.0.0.1:7373  RPC address of the Serf agent.
   -rpc-auth=""              RPC auth token of the Serf agent.
+  -prune					Remove agent forcibly from list of members
```
rboyer: Your whitespace is off.

serf/serf.go
```diff
@@ -785,11 +785,12 @@ func (s *Serf) Members() []Member {
 // immediately, instead of waiting for the reaper to eventually reclaim it.
 // This also has the effect that Serf will no longer attempt to reconnect
 // to this node.
-func (s *Serf) RemoveFailedNode(node string) error {
+func (s *Serf) RemoveFailedNode(node string, prune bool) error {
```
rboyer: Same comment here about breaking the arity of this function. I know this will be a breaking change in nomad.

serf/serf_test.go
```diff
@@ -836,6 +836,68 @@ func TestSerfRemoveFailedNode(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	s3, err := Create(s3Config)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	defer s1.Shutdown()
```
rboyer: You should interlace these after the if/err checks above. That way if `Create` for s3 fails you don't leak the running server for s1 and s2.

serf/serf.go
s-christoff: I think we could totally swap `ForgetNode` and the CacheLock/Delete and it would be easier to read. This function `handlePrune` is a gutted out version of `Reap()`, which I copied the most important parts from - so we do the same pattern there as well. 
As far as out of order alive messages, hopefully the time on the messages will make them easy to refute?

client/rpc_client.go
RobbieMcKinstry: @s-christoff Change this to take a variadic argument of type `Option` instead. That will keep the change in signature from breaking downstream users.

serf/serf_test.go
s-christoff: I'm down for this as it may make it easier to do some test cleanup in the future. How would I go about it? import `"github.com/hashicorp/consul/sdk/testutil/retry"` or is there a way I put it in `"github.com/hashicorp/serf/testutil"`
🤔

serf/serf_test.go
s-christoff: I updated this, can you check to see if what I did matched what you wanted?

client/rpc_client.go
```diff
@@ -193,7 +193,21 @@ func (c *RPCClient) ForceLeave(node string) error {
...
 	}
 	req := forceLeaveRequest{
-		Node: node,
+		Node:  node,
+		Prune: false,
+	}
+	return c.genericRPC(&header, &req, nil)
+}
+
+// If the prune flag is set, remove the node entirely
```
rboyer: The method comment should start like `// ForceLeavePrune ...`

cmd/serf/command/agent/agent.go
```diff
@@ -182,6 +182,17 @@ func (a *Agent) ForceLeave(node string) error {
 	return err
 }
 
+//ForceLeavePrune completely removes a failed node from the
```
rboyer:
```suggestion
// ForceLeavePrune completely removes a failed node from the
```

cmd/serf/command/agent/agent.go
```diff
@@ -182,6 +182,17 @@ func (a *Agent) ForceLeave(node string) error {
...
+//member list entirely
```
rboyer:
```suggestion
// member list entirely
```

cmd/serf/command/agent/agent.go
```diff
@@ -182,6 +182,17 @@ func (a *Agent) ForceLeave(node string) error {
...
+func (a *Agent) ForceLeavePrune(node string) error {
+	a.logger.Printf("[INFO] agent: Force leaving node: %s", node)
```
rboyer: This log message should probably indicate that it's being pruned.

cmd/serf/command/agent/agent.go
```diff
@@ -182,6 +182,17 @@ func (a *Agent) ForceLeave(node string) error {
...
+	err := a.serf.RemoveFailedNodePrune(node)
+	if err != nil {
+		a.logger.Printf("[WARN] agent: failed to remove node: %v", err)
```
rboyer: This error should also mention pruning.

cmd/serf/command/agent/ipc.go
```diff
@@ -605,6 +606,17 @@ func (i *AgentIPC) handleForceLeave(client *IPCClient, seq uint64) error {
...
+	if req.Prune {
+		err := i.agent.ForceLeavePrune(req.Node)
+
+		// Respond
+		resp := responseHeader{
+			Seq:   seq,
+			Error: errToString(err),
+		}
+		return client.Send(&resp, nil)
+	}
```
rboyer: Probably for clarity the remainder of this method should be in an `} else {` block since this flag implies an either/or execution.

cmd/serf/command/force_leave.go
```diff
@@ -33,14 +36,24 @@ func (c *ForceLeaveCommand) Run(args []string) int {
 		return 1
 	}
 
-	client, err := RPCClient(*rpcAddr, *rpcAuth)
+	RPCClient, err := RPCClient(*rpcAddr, *rpcAuth)
```
rboyer: Was there any particular reason for renaming this variable?

serf/serf.go
```diff
@@ -781,15 +781,33 @@ func (s *Serf) Members() []Member {
 	return members
 }
 
-// RemoveFailedNode forcibly removes a failed node from the cluster
+//RemoveFailedNode is a backwards compatabile form
```
rboyer:
```suggestion
// RemoveFailedNode is a backwards compatabile form
```

serf/serf.go
```diff
@@ -781,15 +781,33 @@ func (s *Serf) Members() []Member {
...
+	return nil
+
+}
+
+// ForceLeave forcibly removes a failed node from the cluster
 // immediately, instead of waiting for the reaper to eventually reclaim it.
 // This also has the effect that Serf will no longer attempt to reconnect
 // to this node.
-func (s *Serf) RemoveFailedNode(node string) error {
+func (s *Serf) ForceLeave(node string, prune bool) error {
```
rboyer: Do you need to make this method exported, or can it be named `forceLeave()` and left unexported?

serf/serf.go
```diff
@@ -781,15 +781,33 @@ func (s *Serf) Members() []Member {
 	return members
 }
 
-// RemoveFailedNode forcibly removes a failed node from the cluster
+//RemoveFailedNode is a backwards compatabile form
+// of ForceLeave
+func (s *Serf) RemoveFailedNode(node string) error {
+	if err := s.ForceLeave(node, false); err != nil {
```
rboyer: Since it's just a straight method wrap, you can do this in one line:
```
return s.ForceLeave(node, false)
```
and similar for RemoveFailedNodePrune

serf/serf.go
```diff
@@ -1438,6 +1470,33 @@ func (s *Serf) resolveNodeConflict() {
 	}
 }
 
+// handlePrune is a simplified version of Reap() which only runs when the prune
+//flag is set
```
rboyer:
```suggestion
// flag is set
```

cmd/serf/command/force_leave.go
s-christoff: [Yes, and no](x.co/yesandno) - it was for a different approach but I figured it couldn't _hurt_ to keep it renamed for clarity? Easy to undo if you think it's unnecessary.

serf/serf_test.go
rboyer: You can't have serf depend on consul because that would cause a not fun circular dependency. You could copy just the `retry` package as `github.com/hashicorp/serf/internal/retry` so that serf could use it, but nobody else could.

serf/serf_test.go
rboyer: :+1:

cmd/serf/command/force_leave.go
rboyer: If you want to rename it, please don't have it start with an uppercase letter as that implies that it is exported, which is weird because it's a local variable. Probably just `rpcClient` then.

cmd/serf/command/agent/rpc_client_test.go
```diff
@@ -112,6 +112,59 @@ WAIT:
...
+	client, a1, ipc := testRPCClient(t)
+	a2 := testAgent(nil)
+	defer ipc.Shutdown()
+	defer client.Close()
+	defer a1.Shutdown()
+	defer a2.Shutdown()
+
+	if err := a1.Start(); err != nil {
+		t.Fatalf("err: %s", err)
+	}
```
rboyer: I think the `defer a1.Shutdown()` should be moved to after this startup block. I have no idea if you can shutdown a client that didn't start.

cmd/serf/command/agent/rpc_client_test.go
```diff
@@ -112,6 +112,59 @@ WAIT:
...
+
+	if err := a2.Start(); err != nil {
+		t.Fatalf("err: %s", err)
+	}
```
rboyer: I think the `defer a2.Shutdown()` should be moved to after this startup block.

serf/serf.go
rboyer: @banks the ForgetNode/Lock/delete/Unlock sequence already happens in `reap()`. So if there's a race, it's an old race.

serf/serf.go
s-christoff: I can make it unexported!

serf/serf.go
rboyer: It might be clearer if you had both reap() and handlePrune() share the same block of code:
```
func (s *Serf) removeMember(m *memberState, prune bool) {
	// Delete from members
	delete(s.members, m.Name)

	// Tell the coordinate client the node has gone away and delete
	// its cached coordinates.
	if !s.config.DisableCoordinates {
		s.coordClient.ForgetNode(m.Name)

		s.coordCacheLock.Lock()
		delete(s.coordCache, m.Name)
		s.coordCacheLock.Unlock()
	}

	// Send an event along
	if prune {
		s.logger.Printf("[INFO] serf: EventMemberReap (forced): %s", m.Name)
	} else {
		s.logger.Printf("[INFO] serf: EventMemberReap: %s", m.Name)
	}
	if s.config.EventCh != nil {
		s.config.EventCh <- MemberEvent{
			Type:    EventMemberReap,
			Members: []Member{m.Member},
		}
	}
}
```

Then `reap()` can call `s.removeMember(m, false)` and `handlePrune` can call `s.removeMember(m, true)`

PR #580, opened 2019-09-24
Update ForceLeave Prune

s-christoff:
ForceLeave previously didn't take into consideration what occurs during `StatusLeaving` and `StatusAlive`. The `handlePrune()` function was created to allow for different statuses to prune nodes.

### DIFF
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -1136,13 +1136,7 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
 
 		return true
 
-	case StatusLeaving:
-		if leaveMsg.Prune {
-			s.handlePrune(member)
-		}
-		return true
-
-	case StatusLeft:
+	case StatusLeaving, StatusLeft:
 		if leaveMsg.Prune {
 			s.handlePrune(member)
 		}
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -1135,7 +1135,11 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
 		}
 
 		return true
-
+	case StatusLeft:
+		if leaveMsg.Prune {
+			s.handlePrune(member)
+		}
+		return true
 	default:
 		return false
 	}
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -1102,6 +1102,10 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
 	case StatusAlive:
 		member.Status = StatusLeaving
 		member.statusLTime = leaveMsg.LTime
+
+		if leaveMsg.Prune {
+			s.handlePrune(member)
+		}
 		return true
 	case StatusFailed:
 		member.Status = StatusLeft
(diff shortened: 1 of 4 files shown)


serf/serf.go
```diff
@@ -1127,26 +1131,29 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
...
 		if leaveMsg.Prune {
-			s.logger.Printf("[INFO] serf: EventMemberReap (forced): %s %s", member.Name, member.Member.Addr)
-			s.leftMembers = removeOldMember(s.leftMembers, member.Name)
-			s.eraseNode(member)
+			s.handlePrune(member)
 		}
 
 		return true
 
-	case StatusLeft:
```
rboyer: Did you intentionally remove handling for StatusLeft?

serf/serf.go
```diff
@@ -1127,26 +1131,29 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
...
-
 	default:
 		return false
 	}
 }
 
+// handlePrune waits for nodes that are leaving and then forcibly
+// erases a member from the list of members
+func (s *Serf) handlePrune(member *memberState) {
+	if member.Status == StatusLeaving {
```
rboyer: This will only be called with `StatusAlive` and `StatusLeft` values so this is dead code?

serf/serf.go
```diff
@@ -1127,26 +1131,29 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
...
+		time.Sleep(s.config.BroadcastTimeout + s.config.LeavePropagateDelay)
+	}
+
+	s.logger.Printf("[INFO] serf: EventMemberReap (forced): %s %s", member.Name, member.Member.Addr)
+	s.leftMembers = removeOldMember(s.leftMembers, member.Name)
```
rboyer: Will a node always be in the `leftMembers` list when you invoke `handlePrune` or do you need to handle it being in the `failedMembers` list?

rboyer:
I think this probably warrants a test or two.

serf/serf.go
s-christoff: Nope :(

serf/serf.go
```diff
@@ -1127,26 +1131,44 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
...
-			s.logger.Printf("[INFO] serf: EventMemberReap (forced): %s %s", member.Name, member.Member.Addr)
-			s.leftMembers = removeOldMember(s.leftMembers, member.Name)
-			s.eraseNode(member)
+			s.handlePrune(member)
 		}
 
 		return true
 
-	case StatusLeft:
+	case StatusLeaving:
```
rboyer: You don't have to do this, but I'm just noting that since these two cases have the same bodies you could change it to:
```
case StatusLeaving, StatusLeft:
   ...common body...
```

serf/serf_test.go
```diff
@@ -517,6 +539,179 @@ func TestSerf_leaveRejoinDifferentRole(t *testing.T) {
...
+	defer s2.Shutdown()
+
+	_, err = s1.Join([]string{s2Config.MemberlistConfig.BindAddr}, false)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	// s1 should be alive when we call the force leave
+	// but not refute it
+	s2.forceLeave(s1.config.NodeName, true)
```
rboyer: In the other two variations of this test you use `s2` as the node under test (the one you are leaving) and `s1` as the unchanged node, but in this test you flipped the roles around.  For ease of reading you may want to reverse the roles in this test so all of them read the same.

s-christoff: fixed!

PR #584, opened 2019-10-10
tests: normalize and unflake a bunch of tests

rboyer:
Also:
- sync some makefile targets back from consul
- reroute test log output through testing.T.Logf
- include memberlist logs with serf logs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,8 @@ jobs:
           keys:
             - serf-modcache-v1-{{ checksum "go.mod" }}
 
+      - run: sudo apt-get update && sudo apt-get install -y rsyslog
+      - run: sudo service rsyslog start
       # run go tests with gotestsum
       - run: |
           PACKAGE_NAMES=$(go list ./...)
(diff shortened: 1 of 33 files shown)


rboyer:
The bulk of the test changes boil down to:

- treating the bind ips used for the tests as a scarce resource and keeping track of when they go in and out of scope so that you can't accidentally interlock two tests
- cutting back on use of `testutil.Yield()` in favor of `retry.Run` waiting on a condition

testutil/addr.go
```diff
@@ -1,23 +1,76 @@
...
+	"fmt"
 	"net"
+	"os"
 	"sync"
+	"time"
 )
 
-var bindLock sync.Mutex
-var bindNum byte = 10
+var (
```
rboyer: The changes here are an `ip` variation of the thing I recently fixed in consul's `sdk/freeport` library.

.circleci/config.yml
```diff
@@ -51,6 +51,8 @@ jobs:
           keys:
             - serf-modcache-v1-{{ checksum "go.mod" }}
 
+      - run: sudo apt-get update && sudo apt-get install -y rsyslog
```
rboyer: Lifted this from consul's circleci configs to get a syslog-related serf test to actually pass.

PR #590, opened 2019-12-04
plumbing changes necessary to allow for gossip through an intermediary network hop

rboyer:
Dependent upon: https://github.com/hashicorp/memberlist/pull/210


serf/messages.go
```diff
@@ -144,22 +145,33 @@ func encodeMessage(t messageType, msg interface{}) ([]byte, error) {
 // relayHeader is used to store the end destination of a relayed message
 type relayHeader struct {
 	DestAddr net.UDPAddr
+	DestName string
```
banks: 🤔 I guess this is backwards compatible for older Serf members they just won't include a name. Would that cause issues if for example only Consul servers were updated and used names but client agents weren't and then Serf queries were used?

rboyer: _(serf-only)_

Since all of the fields are optional/additive you should be able to mix old and new serf instances in the same gossip pool without anything tragic happening.

If either `config.MemberlistConfig.RequireNodeNames==true` or the provided `memberlist.Transport` somehow _required_ a non empty node name that's when the tragedy comes into play.

_(consul)_

In the context of https://github.com/hashicorp/consul/pull/6884 we will only be setting `RequireNodeNames=true` and overriding the Transport with one that requires node names when federation via mesh gateways is enabled. The guide (TBD) will call out that you need to first complete the upgrade of all consul servers _everywhere_ and the clients co-located with mesh gateways as a precondition _before_ toggling the newfangled federation on to avoid just such a problem.

PR #593, opened 2019-12-19
fix missing entries in the vendor directory

rboyer:
Restores some stuff that was removed in #592

```
$ rm -rf vendor
$ go mod tidy
$ go mod vendor
$ git add -u
$ git add vendor
```


notnoop:
Thanks - weird, I ran `go mod vendor` but without `go mod tidy`, so I assumed those packages were no longer needed.  How would I have detected the issue better?

rboyer:
Ah. Consul is still on the `1.12.x` series so that's what I ran it with.

PR #595, opened 2020-02-15
Implement Primary Key Retrieval Command

vladfau:
Hi guys, 

Initially, I had a requirement for Consul to obtain the currently used primary encryption key, but it looks like I've gotta start from Serf first. Please let me know what do you think about this approach and do you find it useful at all? I can add tests for UI part later on if you consider it a right way to achieve my original task.

### DIFF
--- a/client/const.go
+++ b/client/const.go
@@ -27,6 +27,7 @@ const (
 	useKeyCommand          = "use-key"
 	removeKeyCommand       = "remove-key"
 	listKeysCommand        = "list-keys"
+	getPrimaryKeyCommand   = "get-primary-key"
 	tagsCommand            = "tags"
 	queryCommand           = "query"
 	respondCommand         = "respond"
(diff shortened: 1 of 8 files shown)


i0rek:
@vladfau thank you very much for your contribution. When I read the headline I expected that the code would return the primary key of the local node, but it is asking every node for their primary keys. 

I am curious about the use case for such an operation, could you explain what you would want to do with this operation?

vladfau:
@i0rek, thanks for replying to my PR!

My idea here was simple – mimic the behavior of `ListKeys()`, but only retrieve the primary one only as I was unable to identify the way how to request  this key from the current member. Also, I do not think that we can avoid querying all the members for CLI-based calls, when the caller is not member of the cluster (basing on what I understand in Serf).

kyhavlov:
Hi @vladfau - you mentioned the motivation here was to obtain the current primary key in Consul - are you looking for something similar to the existing `-list` or `-list-primary` flags under the `keyring` command? https://www.consul.io/commands/keyring#list-primary

vladfau:
Hello @kyhavlov, 

I am no longer working with this project. I can say that if -list-primary returns the current primary key and it is always returns the active primary one without any permutations, this PR can be closed

PR #599, opened 2020-04-06
move version command into a package

alvin-huang:
This moves the logic of calculating the version into its own package. This is to better standardize across the various libraries and ecosystem tools so we can share the same build logic and pipeline. The build logic will look to modify `Version` and `VersionPrerelease` within `version/version.go`


version/version.go
```diff
@@ -0,0 +1,42 @@
...
+// GetHumanVersion composes the parts of the version in a way that's suitable
+// for displaying to humans.
+func GetHumanVersion() string {
+	version := Version
+	release := VersionPrerelease
+
+	if release != "" {
+		if !strings.HasSuffix(version, "-"+release) {
+			// if we tagged a prerelease version then the release is in the version already
+			version += fmt.Sprintf("-%s", release)
```
alvin-huang, line 33: @mkeeler this will change the current delimiter for prerelease versions from `.` (https://github.com/hashicorp/serf/pull/599/files#diff-f249ac3c4d1639887402e5e0d7326f0bL29) to `-` to be consistent with other projects.

PR #606, opened 2020-07-20
Avoid issue where two unique leave events for the same node could lead to infinite rebroadcast storms

rboyer:
Fixes this scenario:

When two leave events for the same node (with different ltimes) are
received by two or more additional nodes without any intermediate state
transitions it will lead to them setting up an infinite gossip echo
chamber rebroadcasting the original node's leave.

When in this situation the only way to recover would be to either:

- Bring the node back so it can send out a fresh "alive" message.

- Use force-leave prune to purge the node's existence from the cluster.

There are two known ways to arrive in this situation:

1. Have a node leave the cluster and then from a differnet node initiate
   a force-leave. This is the easiest way to reproduce.

2. Have a node leave, rejoin, and leave again in quick succession. This
   requires winning a lot of races in just the right way to have gossip
   from the first leave still rebroadcasting naturally WHILE the second
   round of leaves is rebroadcasting AND to have those rounds of gossip
   arrive at the target nodes before they see the rebroadcast of the
   intervening REJOIN.

The fix is pretty simple. When we witness a node leave event, before we
rebroadcast we need to update the lamport time associated with that node
so that the ltime check will cause the same event to be processed again
and again.

Fixes https://github.com/hashicorp/consul/issues/7960
Fixes https://github.com/hashicorp/consul/issues/8179

### DIFF
--- a/serf/config.go
+++ b/serf/config.go
@@ -247,10 +247,12 @@ type Config struct {
 	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
 	UserEventSizeLimit int
 
-	// MessageDropper is a callback used for selectively ignoring inbound
+	// messageDropper is a callback used for selectively ignoring inbound
 	// gossip messages. This should only be used in unit tests needing careful
 	// control over sequencing of gossip arrival
-	MessageDropper func(typ messageType) bool
+	//
+	// WARNING: this should ONLY be used in tests
+	messageDropper func(typ messageType) bool
 }
 
 // Init allocates the subdata structures
@@ -258,8 +260,8 @@ func (c *Config) Init() {
 	if c.Tags == nil {
 		c.Tags = make(map[string]string)
 	}
-	if c.MessageDropper == nil {
-		c.MessageDropper = func(typ messageType) bool {
+	if c.messageDropper == nil {
+		c.messageDropper = func(typ messageType) bool {
 			return false
 		}
 	}
--- a/serf/config.go
+++ b/serf/config.go
@@ -246,12 +246,22 @@ type Config struct {
 	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
 	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
 	UserEventSizeLimit int
+
+	// MessageDropper is a callback used for selectively ignoring inbound
+	// gossip messages. This should only be used in unit tests needing careful
+	// control over sequencing of gossip arrival
+	MessageDropper func(typ messageType) bool
 }
 
 // Init allocates the subdata structures
 func (c *Config) Init() {
 	if c.Tags == nil {
 		c.Tags = make(map[string]string)
+	}
+	if c.MessageDropper == nil {
+		c.MessageDropper = func(typ messageType) bool {
+			return false
+		}
 	}
 }
 
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -907,7 +907,7 @@ func (s *Serf) handleNodeJoin(n *memberlist.Node) {
 	s.memberLock.Lock()
 	defer s.memberLock.Unlock()
 
-	if s.config.MessageDropper(messageJoinType) {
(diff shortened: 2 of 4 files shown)


serf/serf.go
```diff
@@ -1098,6 +1102,10 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
 		return false
 	}
 
+	// Always set the lamport time so that if we retransmit below it won't echo
+	// around forever!
+	member.statusLTime = leaveMsg.LTime
+
```
s-christoff, line 1108: Should we redo this in each case if we are going to be setting it outside?

serf/config.go
```diff
@@ -246,13 +246,23 @@ type Config struct {
 	// UserEventSizeLimit is maximum byte size limit of user event `name` + `payload` in bytes.
 	// It's optimal to be relatively small, since it's going to be gossiped through the cluster.
 	UserEventSizeLimit int
+
+	// MessageDropper is a callback used for selectively ignoring inbound
+	// gossip messages. This should only be used in unit tests needing careful
+	// control over sequencing of gossip arrival
+	MessageDropper func(typ messageType) bool
```
mkeeler, line 253:
```suggestion
	messageDropper func(typ messageType) bool
```

Making this unexported would prevent it from being used outside of the library and our tests.

rboyer: I thought about this, but it would also make it harder to "copy" the `Config` struct, right?

rboyer: Apparently the thing I thought was a problem isn't.

serf/serf.go
rboyer: done

serf/serf.go
```diff
@@ -1098,19 +1102,21 @@ func (s *Serf) handleNodeLeaveIntent(leaveMsg *messageLeave) bool {
 		return false
 	}
 
+	// Always set the lamport time so that if we retransmit below it won't echo
+	// around forever!
```
banks, line 1106: @rboyer we talked about making this comment more extensive:

 - pointing out that the name is a little misleading since the status might not actually change but that the sideeffect or preventing rebroadcasts in the laast case is still needed.
 - Perhaps linking to an issue or PR that described the problems caused by this for future us trying to work out why we moved it here!

I realise this is merged already but perhaps we could update the comment on master separately while context is fresh - might need to be after we have provided a clearer writeup of the issue somewhere public like the Consul linked issue?

serf/serf_test.go
```diff
@@ -276,6 +278,156 @@ func TestSerf_eventsLeave(t *testing.T) {
...
+		case messageJoinType, messagePushPullType:
+			return atomic.LoadUint32(&s3DropJoins) == 1
+		default:
+			return false
+		}
+	}
+
+	s4Config := testConfigLocal(t, ip4)
+	// Allow s4 to drop joins in the future.
+	var s4DropJoins uint32
```
banks, line 331: We could use the same uint32 and dropper func for both s3 and s4 since we switch them both together anyway?

serf/serf_test.go
```diff
@@ -276,6 +278,156 @@ func TestSerf_eventsLeave(t *testing.T) {
...
+	defer s2.Shutdown()
+
+	_, err = s2.Join([]string{s1Config.NodeName + "/" + s1Config.MemberlistConfig.BindAddr}, false)
+	if err != nil {
+		t.Fatalf("err: %v", err)
+	}
+
+	waitUntilNumNodes(t, 4, s1, s2, s3, s4)
+
+	// Now leave a second time but before s3 saw the rejoin (due to the gate)
```
banks, line 411: nit: `s/s3/s3 and s4/`

banks:
Great job @rboyer! This turned out much cleaner than my original hack!

serf/serf.go
rboyer: new PR with this feedback: https://github.com/hashicorp/serf/pull/608

PR #609, opened 2020-07-24
Fix getting VERSION from version.go in makefile

bayandin:
Currently `make dist` fails because it can't find file `cmd/serf/version.go`:

```
$ make dist
awk: can't open file cmd/serf/version.go
 source line number 1
Please specify a version.
make: *** [dist] Error 1
```


markan (approved):
Looks good to me.

PR #611, opened 2020-08-17
add primary key to keylist response

i0rek:
This PR modifies the key list response and adds a field for the primary key of a node. They will be accumulated similarly to the other keys.

### DIFF
--- a/serf/internal_query_test.go
+++ b/serf/internal_query_test.go
@@ -131,7 +131,7 @@ func TestSerfQueries_keyListResponseWithCorrectSize(t *testing.T) {
 		{expected: 0, hasMsg: false, resp: nodeKeyResponse{}},
 		{expected: 1, hasMsg: false, resp: nodeKeyResponse{Keys: []string{"KfCPZAKdgHUOdb202afZfE8EbdZqj4+ReTbfJUkfKsg="}}},
 		// has 50 keys which makes the response bigger than 1024 bytes.
-		{expected: 19, hasMsg: true, resp: nodeKeyResponse{Keys: []string{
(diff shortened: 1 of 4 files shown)


serf/internal_query_test.go
```diff
@@ -131,7 +131,7 @@ func TestSerfQueries_keyListResponseWithCorrectSize(t *testing.T) {
 		{expected: 0, hasMsg: false, resp: nodeKeyResponse{}},
 		{expected: 1, hasMsg: false, resp: nodeKeyResponse{Keys: []string{"KfCPZAKdgHUOdb202afZfE8EbdZqj4+ReTbfJUkfKsg="}}},
 		// has 50 keys which makes the response bigger than 1024 bytes.
-		{expected: 19, hasMsg: true, resp: nodeKeyResponse{Keys: []string{
+		{expected: 18, hasMsg: true, resp: nodeKeyResponse{Keys: []string{
```
i0rek, line 134: This changes because the primary key was added to the response. Which is why less keys fit into the message.

serf/keymanager.go
```diff
@@ -83,6 +87,12 @@ func (k *KeyManager) streamKeyResp(resp *KeyResponse, ch <-chan NodeResponse) {
 			}
 		}
 
+		if _, ok := resp.PrimaryKeys[nodeResponse.PrimaryKey]; !ok {
+			resp.PrimaryKeys[nodeResponse.PrimaryKey] = 1
+		} else {
+			resp.PrimaryKeys[nodeResponse.PrimaryKey]++
+		}
```
mkeeler, line 94:
```suggestion
		resp.PrimaryKeys[nodeResponse.PrimaryKey] += 1
```

This works because if the key doesn't exist the zero value is returned.

PR #612, opened 2020-08-20
Validate NodeNames, Address and Tags fields

s-christoff:
We've been seeing an issue where NodeNames will start to get very long and filled with junk. 
Fields have been added to validate memberinfo in messages are within certain restrictions. 

A configuration flag titled `ValidateNodeNames` has been added to give users the option of restricting nodename length.

The Address field in the memberinfo in messages must adhere to an IP:Port structure.
The Tags field in memberinfo in messages must adhere to the 512 max (which is placed elsewhere too)
ValidateNodeNames restricts NodeNames to alphanumeric and 128 characters.

### DIFF
--- a/cmd/serf/command/agent/config.go
+++ b/cmd/serf/command/agent/config.go
@@ -19,7 +19,6 @@ import (
 
 // This is the default port that we use for Serf communication
 const DefaultBindPort int = 7946
-const MaxNodeName int = 128
 
 // DefaultConfig contains the defaults for configurations.
 func DefaultConfig() *Config {
@@ -358,7 +357,7 @@ func DecodeConfig(r io.Reader) (*Config, error) {
 			return nil, err
 		}
 
-		if len(result.NodeName) > MaxNodeName {
+		if len(result.NodeName) > serf.MaxNodeNameLength {
 			err = fmt.Errorf("NodeName is %v characters. "+
 				"Valid length is between 1 and 128 characters", len(result.NodeName))
 			return nil, err
--- a/cmd/serf/command/agent/config.go
+++ b/cmd/serf/command/agent/config.go
@@ -228,6 +228,11 @@ type Config struct {
 	// 5 seconds.
 	BroadcastTimeoutRaw string        `mapstructure:"broadcast_timeout"`
 	BroadcastTimeout    time.Duration `mapstructure:"-"`
+
+	//TODO(schristoff): better var name? :grimance:
+	//NodeNameValid specifies whether or not nodenames should
+	// be alphanumeric and within 128 characters
+	NodeNameValid bool `mapstructure:NodeNamevalid`
 }
 
 // BindAddrParts returns the parts of the BindAddr that should be
@@ -354,10 +359,10 @@ func DecodeConfig(r io.Reader) (*Config, error) {
 			return nil, err
 		}
 
-		MaxLength := 63
+		MaxLength := 128
 		if len(result.NodeName) > MaxLength {
 			err = fmt.Errorf("NodeName is %v characters. "+
-				"Valid length is between 1 and 63 characters", len(result.NodeName))
+				"Valid length is between 1 and 128 characters", len(result.NodeName))
 			return nil, err
 		}
 	}
@@ -381,9 +386,9 @@ func containsKey(keys []string, key string) bool {
 func MergeConfig(a, b *Config) *Config {
 	var result Config = *a
 
-	// Copy the strings if they're set
 	//TODO(schristoff): do i need to check nodename
 	//validatity here?
+
 	if b.NodeName != "" {
 		result.NodeName = b.NodeName
 	}
--- a/cmd/serf/command/agent/config.go
+++ b/cmd/serf/command/agent/config.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -347,21 +346,6 @@ func DecodeConfig(r io.Reader) (*Config, error) {
 			return nil, err
 		}
 		result.BroadcastTimeout = dur
-	}
-
-	if result.NodeName != "" && result.ValidateNodeNames {
-		var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\\-]+`)
-		if InvalidNameRe.MatchString(result.NodeName) {
-			err = fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
-				"all alpha-numerics and dashes.", result.NodeName)
-			return nil, err
-		}
-
-		if len(result.NodeName) > serf.MaxNodeNameLength {
-			err = fmt.Errorf("NodeName is %v characters. "+
-				"Valid length is between 1 and 128 characters", len(result.NodeName))
-			return nil, err
-		}
 	}
 
 	return &result, nil
--- a/cmd/serf/command/agent/config.go
+++ b/cmd/serf/command/agent/config.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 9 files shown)


cmd/serf/command/agent/config.go
```diff
@@ -227,6 +228,11 @@ type Config struct {
 	// 5 seconds.
 	BroadcastTimeoutRaw string        `mapstructure:"broadcast_timeout"`
 	BroadcastTimeout    time.Duration `mapstructure:"-"`
+
+	//TODO(schristoff): better var name? :grimance:
+	//NodeNameValid specifies whether or not nodenames should
+	// be alphanumeric and within 128 characters
+	NodeNameValid bool `mapstructure:NodeNamevalid`
```
kyhavlov, line 235:
```suggestion
	NodeNameValid bool `mapstructure:"node_name_valid"`
```

cmd/serf/command/agent/config.go
```diff
@@ -344,6 +350,23 @@ func DecodeConfig(r io.Reader) (*Config, error) {
...
+	//TODO(schristoff): this will prolly break some stuff, add config flag for this check?
+	if result.NodeName != "" {
+		var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\\-]+`)
+		if InvalidNameRe.MatchString(result.NodeName) {
+			err = fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
+				"all alpha-numerics and dashes.", result.NodeName)
+			return nil, err
+		}
+
+		MaxLength := 128
```
kyhavlov, line 362: This can be a constant set above, like with `DefaultBindPort`

serf/config.go
```diff
@@ -271,6 +273,16 @@ func (c *Config) Init() {
 // for most of the configurations.
 func DefaultConfig() *Config {
 	hostname, err := os.Hostname()
+	//TODO(schristoff): is it safe to add this here? also should be flag
+	if len(hostname) > 128 {
+		err = fmt.Errorf("Hostname is %v characters. Valid length is between 1 and 128 characters", len(hostname))
+	}
```
kyhavlov, line 279: I think this is ok to validate here - we should use the length constant from above

serf/merge_delegate.go
```diff
@@ -45,5 +56,28 @@ func (m *mergeDelegate) nodeToMember(n *memberlist.Node) *Member {
...
+		return fmt.Errorf("NodeName length is %v characters. Valid length is between "+
+			"1 and 128 characters.", len(n.Name))
+	}
+	if InvalidNameRe.MatchString(n.Name) {
+		return fmt.Errorf("Nodename contains invalid characters %v , Valid characters include "+
+			"all alpha-numerics and dashes", n.Name)
+	}
+
+	if net.ParseIP(string(n.Addr)) == nil {
+		return fmt.Errorf("Address is %v . Must be a valid representation of an IP address. ", n.Addr)
```
kyhavlov, line 76: We should check for `address:port` format; the field is called Addr but stores the port too.

serf/merge_delegate.go
```diff
@@ -15,24 +17,33 @@ type mergeDelegate struct {
 }
 
 func (m *mergeDelegate) NotifyMerge(nodes []*memberlist.Node) error {
+	var err error
```
kyhavlov, line 20: tiny nitpick, but this could go in the for loop since it's only being used in that scope

cmd/serf/command/agent/config.go
```diff
@@ -18,6 +19,7 @@ import (
 
 // This is the default port that we use for Serf communication
 const DefaultBindPort int = 7946
+const MaxNodeName int = 128
```
kyhavlov, line 22: This should probably reference the field in the serf package with the same name instead of defining it again here

serf/config.go
```diff
@@ -14,6 +16,8 @@ import (
 // our own protocol version.
 var ProtocolVersionMap map[uint8]uint8
 
+const MaxNodeName int = 128
```
kyhavlov, line 19: This could be `MaxNodeNameLength` instead

serf/config.go
```diff
@@ -253,6 +257,11 @@ type Config struct {
 	//
 	// WARNING: this should ONLY be used in tests
 	messageDropper func(typ messageType) bool
+
+	//ValidateNodeNames specifies whether or not nodenames should
+	// be alphanumeric and within 128 characters
+	//TODO(schristoff): should this be here?
```
kyhavlov, line 263: I think this is the right place for this setting - this config controls how the serf instance behaves

serf/config.go
```diff
@@ -275,6 +284,18 @@ func DefaultConfig() *Config {
...
+	if len(hostname) > MaxNodeName {
+		//TODO(schristoff): panic seems a little harsh here?
+		panic(fmt.Errorf("NodeName is %v characters. "+
+			"Valid length is between 1 and 128 characters", len(hostname)))
+	}
+	var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\\-]+`)
+	if InvalidNameRe.MatchString(hostname) {
+		panic(fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
+			"all alpha-numerics and dashes.", hostname))
+	}
```
kyhavlov, line 297: I think we have to do this validation in Create(), or else it won't be checked when config gets passed in from Consul. We won't have to check it here and potentially panic then either, since it'll just return an error when trying to create the serf instance.

cmd/serf/command/agent/config.go
```diff
@@ -344,6 +349,21 @@ func DecodeConfig(r io.Reader) (*Config, error) {
...
+			err = fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
+				"all alpha-numerics and dashes.", result.NodeName)
+			return nil, err
+		}
+
+		if len(result.NodeName) > serf.MaxNodeNameLength {
+			err = fmt.Errorf("NodeName is %v characters. "+
+				"Valid length is between 1 and 128 characters", len(result.NodeName))
+			return nil, err
+		}
```
kyhavlov, line 364: This logic for validating the name is getting repeated 3 times - we should pull it into a function for reuse. We can also avoid doing it in this `serf/command/agent` code at all and do it here instead: https://github.com/hashicorp/serf/blob/master/serf/serf.go#L235

`serf.Create` will get called on startup for Consul as well, where the code in this package is only used in the Serf binary.

serf/merge_delegate.go
```diff
@@ -45,5 +56,31 @@ func (m *mergeDelegate) nodeToMember(n *memberlist.Node) *Member {
...
 		DelegateMax: n.DMax,
 		DelegateCur: n.DCur,
+	}, nil
+}
+
+// validateMemberInfo checks that the data we are sending is valid
+func (m *mergeDelegate) validiateMemberInfo(n *memberlist.Node) error {
+	var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\\-]+`)
+	var InvalidIPv4Re = regexp.MustCompile(`([0-9]{1,3}\.){3}[0-9]{1,3}:[0-9]+`)
+	var InvalidIPv6Re = regexp.MustCompile(`([0-9a-f]{1,4}:){7}[0-9a-f]{1,4}`)
```
kyhavlov, line 66: I think instead of using regexes for validating the address/port we could do something like this: 
```
host, port, err := net.SplitHostPort(n.Addr)
if err != nil {
   return err
}

ip := net.ParseIP(host)
if ip == nil || (ip.To4() == nil && ip.To16() == nil) {
    fmt.Printf("%v is not a valid IPv4 or 1Pv6 address\n", ip)
    return false
}

p, err := strconv.Atoi(port)
if err != nil {
    return err
}
if p < 0 || p > 65535 {
    return fmt.Errorf("port must be a valid number from 0-65535")
} 
```

This way we can use the go stdlib stuff to validate the IP address format.

serf/serf_test.go
```diff
@@ -2971,3 +2971,31 @@ func waitUntilIntentQueueLen(t *testing.T, desiredLen int, serfs ...*Serf) {
 		}
 	})
 }
+
+func TestAgentCreate_ValidateNodeName(t *testing.T) {
```
kyhavlov, line 2975: This should probably be `TestSerf_ValidateNodeName`

cmd/serf/command/agent/config.go
```diff
@@ -363,7 +367,9 @@ func containsKey(keys []string, key string) bool {
 func MergeConfig(a, b *Config) *Config {
 	var result Config = *a
 
-	// Copy the strings if they're set
+	//TODO(schristoff): do i need to check nodename
```
kyhavlov, line 370: I think we can remove this todo (since we're verifying the final config that gets passed to Create() anyway)

serf/serf.go
```diff
@@ -1884,3 +1890,20 @@ func (s *Serf) NumNodes() (numNodes int) {
 
 	return numNodes
 }
+
+// ValidateNodeNames checks the ValidateNodeNames flag to be true
+// then runs validations
+func (s *Serf) ValidateNodeNames() error {
+	if s.config.ValidateNodeNames {
+		var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\-\.\\]+`)
```
kyhavlov, line 1898: I think the regexp should be `[^A-Za-z0-9\-\.]+` - right now it's allowing `\` characters

serf/serf.go
```diff
@@ -1884,3 +1890,20 @@ func (s *Serf) NumNodes() (numNodes int) {
...
+		if InvalidNameRe.MatchString(s.config.NodeName) {
+			return fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
+				"all alpha-numerics and dashes.", s.config.NodeName)
```
kyhavlov, line 1901: We should specify alphanumerics, dashes and `.`s. Same with the comments for `ValidateNodeNames` above.

PR #614, opened 2020-09-30
Add the ability for per-node reconnect timeouts to be specified by a Serf user

mkeeler:
There are actually 3 commits in this PR:

* The first commit is to make the golangci linter happy with simplifying some range expressions.
* The second commit fixes an issue recently introduced that would cause all messages to be rejected for invalid reasons.
* The third commit is the real work of this PR which allows an application using serf to override the reconnect timeout on a per-node basis. There is an interface they can optionally provide and serf will use its only method to request the per-node timeout. If no interface is provided then the normal reconnect timeout is used.

TODOS:
- [x] Tests

### DIFF
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -1894,15 +1894,19 @@ func (s *Serf) NumNodes() (numNodes int) {
 // ValidateNodeNames verifies the NodeName contains
 // only alphanumeric, -, or . and is under 128 chracters
 func (s *Serf) ValidateNodeNames() error {
+	return s.validateNodeName(s.config.NodeName)
+}
+
+func (s *Serf) validateNodeName(name string) error {
 	if s.config.ValidateNodeNames {
 		var InvalidNameRe = regexp.MustCompile(`[^A-Za-z0-9\-\.]+`)
-		if InvalidNameRe.MatchString(s.config.NodeName) {
-			return fmt.Errorf("NodeName contains invalid characters %v , Valid characters include "+
-				"all alpha-numerics and dashes and '.' ", s.config.NodeName)
+		if InvalidNameRe.MatchString(name) {
+			return fmt.Errorf("Node name contains invalid characters %v , Valid characters include "+
+				"all alpha-numerics and dashes and '.' ", name)
 		}
-		if len(s.config.NodeName) > MaxNodeNameLength {
-			return fmt.Errorf("NodeName is %v characters. "+
-				"Valid length is between 1 and 128 characters", len(s.config.NodeName))
+		if len(name) > MaxNodeNameLength {
+			return fmt.Errorf("Node name is %v characters. "+
+				"Valid length is between 1 and 128 characters", len(name))
 		}
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


s-christoff (approved):
Looks good to me 👍

serf/serf.go
```diff
@@ -1577,6 +1583,10 @@ func (s *Serf) reap(old []*memberState, now time.Time, timeout time.Duration) []
 	for i := 0; i < n; i++ {
 		m := old[i]
 
+		if s.config.ReconnectTimeoutOverride != nil {
+			timeout = s.config.ReconnectTimeoutOverride.ReconnectTimeout(&m.Member, timeout)
```
rboyer, line 1587: This loops over several members, but the original `timeout` from the args is mutated here on _each_ loop, rather than doing something like `memberTimeout := timeout` just before this conditional, and then passing `memberTimeout` into the `ReconnectTimeout` inputs and assigning to the output.

serf/serf.go
```diff
@@ -1577,6 +1583,10 @@ func (s *Serf) reap(old []*memberState, now time.Time, timeout time.Duration) []
...
+		}
+
 		// Skip if the timeout is not yet reached
 		if now.Sub(m.leaveTime) <= timeout {
```
rboyer, line 1591: If you use the renamed timeout value, make sure you rename it here as well.

serf/serf_test.go
```diff
@@ -2999,3 +2999,68 @@ func TestSerf_ValidateNodeName(t *testing.T) {
...
+	waitUntilNumNodes(t, 1, s1)
+
+	// Since s2 shutdown, we check the events to make sure we got failures.
+	testEvents(t, eventCh, s2Config.NodeName,
+		[]EventType{EventMemberJoin, EventMemberFailed, EventMemberReap})
+
+	if !override.called {
+		t.Fatalf("The reconnect override was not used")
+	}
+}
```
rboyer, line 3066: Should you add a reap test that uses artificially _inflated_ timeouts for a node so you can prove that some nodes get reaped as you'd expect and others linger as per the inputs? That might be easier to assert without awful sleeps slowing down the test than doing it the other way around.

serf/serf.go
mkeeler: Good catch

serf/serf_test.go
mkeeler: I actually think inflating the timeouts would make the test take longer. You still have to wait for memberlist to do its business of putting it into a suspect state then doing fallback pings etc. Then we would need to have an arbitrary sleep for a second or 2 to ensure that the node doesn't get reaped.

The test as written executes in about 2s with 1s between when we initiate the shutdown of the node to when we detect that its gone. To me having this test run for 2s seems fine and there are no arbitrary sleeps (just one retry.Run loop for a max of 7s but that should only execute for at most 1s until the reap happens - unless something breaks functionality and causes it not to be reaped)

rboyer (approved):
:shipit:

PR #628, opened 2021-03-25
coordinates.html.markdown: simplify duration calc

peterbourgon:


eculver (approved):
This looks harmless. I'm not sure why we wrote it with the assignment before. Thanks for the contribution!

PR #630, opened 2021-04-26
Fix mdns with CVE-2019-19794

### LINKED ISSUE #629: CVE-2019-19794 on last release by repos hashicorp/mdns
Hello,

The last release has a dependancies with `github.com/miekg/dns v1.0.14` (by dependancie `github.com/hashicorp/mdns`) who's concerned by the CVE-2019-19794 :

https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMMIEKGDNS-537825
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-19794
Could you plan a new release to fix this dependance ?

(and do same will all other repository of hashicorps who's including this repos)

Thanks in advance.

adzeitor:
Current version of hashicorp/mdns using vulnerable
package miekg/dns v1.0.14.

> The miekg Go DNS package before 1.1.25, as used in CoreDNS
> before 1.6.6 and other products, improperly generates
> random numbers because math/rand is used. The TXID becomes
> predictable, leading to response forgeries.

See https://nvd.nist.gov/vuln/detail/CVE-2019-19794

Closes #629


dnephin:
Thank you for the PR! I rebased the PR to fix the conflict, but now `TestCommandRun_mDNS` seems to be failing. It seems like the test failure may be related. It does pass locally, but the failure seems related to ipv6 which I believe is not supported on CircleCI. We may need to adjust the test to only use ipv4, but I haven't looked into it much.

dnephin:
The test failures looks to be related to this change: https://github.com/hashicorp/mdns/pull/73

The underlying error is hidden in the test output, because the tests are using `MockUI`, but when I changed that to `BasicUI` I saw this:

> [ERR] agent.mdns: Failed to poll for new hosts: write udp6 [::]:59553->[ff02::fb]:5353: sendto: cannot assign requested address

What I'm not sure about is why `ipv6UnicastConn.WriteToUDP` is being called. I would have expected `ipv6UnicastConn` to be nil, since we can see the bind failed with "setsockopt: no such device".

Maybe it's only the IPv6 multicast that failed to bind, and `ipv6UnicastConn` is indeed not nil, but still fails to write anyway.

dnephin:
I forgot to include the links to the relevant lines of code. It seems like this `ListenUDP` operation succeeds: https://github.com/hashicorp/mdns/blob/v1.0.4/client.go#L118-L121 , because the return value is not-nil

but later on, this operation: https://github.com/hashicorp/mdns/blob/v1.0.4/client.go#L316-L321 fails with 

> write udp6 [::]:59553->[ff02::fb]:5353: sendto: cannot assign requested address

I'm not sure why it only fails when `WriteToUDP`, and not earlier. Or how to work around that.

dnephin:
Thank you for the PR! The test passes locally, so it is indeed just a problem with the CI environment. We decided to keep skipping the test in CI. I pushed that change to this PR, so we're all set to merge.

PR #632, opened 2021-06-09
ci: use latest Go version and fix a flaky test

dnephin:
Also remove module cache, because downloading from the module proxy is faster than using the cache.


dnephin:
`TestSnapshotter_forceCompact` seems to flake a lot as well, but it does pass locally.

rboyer (approved):
LGTM

PR #633, opened 2021-06-09
Remove vendor directory

dnephin:
Branched from #632

The vendor directory is not up-to-date. Running `go mod vendor` shows files
are updated. Since this package has been using Go modules for a while now we
no longer need the vendor directory.


rboyer:
There's one teeny reference to `vendor` in `GNUMakefile` you might want to remove while you're at it:

```
GNUmakefile:GOFILES ?= $(shell go list ./... | grep -v /vendor/)
```

dnephin:
Good call, removed it from the makefile

rboyer (approved):
LGTM

PR #635, opened 2021-09-21
Add source node getter

femioladipo:
Users of the `serf` library may need to perform actions out of serf regarding the node initiating a query, therefore being able to access this information will be useful.


dnephin (approved):
LGTM, thank you!

PR #637, opened 2021-10-01
fix _simulator.js number of nodes alert

### LINKED ISSUE #636: Website Simulator Alert for Number of Nodes Check is Off-by-One
Hi All!

Simple UI bug: in the [Serf Convergence Simulator](https://www.serf.io/docs/internals/simulator.html), when you specify the integer number of nodes as `1`, an alert pops up saying `You must specify at least 1 node!`, which is simply off by one.

This HTML element: `<input type="text" id="nodes" value="30" class="simulator-input">`

This script logic (line `11378`):
```javascript
return isNaN(t) || 1 >= t ? void alert("Must have at least one node!") : (simulator.nodes = t, void simulator.draw())
```
I believe simply needs to become:
```javascript
return isNaN(t) || 2 > t ? void alert("Must have at least one node!") : (simulator.nodes = t, void simulator.draw())
```

Tada! simple :)

justinTM:
Fixes #636

### DIFF
--- a/website/source/assets/javascripts/_simulator.js
+++ b/website/source/assets/javascripts/_simulator.js
@@ -95,7 +95,7 @@ var SerfSimulator = {
 
     this.$nodes.on("change", function() {
       var nodes = Number(this.value);
-      if (isNaN(nodes) || nodes <= 1) {
+      if (isNaN(nodes) || nodes < 1) {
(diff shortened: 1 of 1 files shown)


dnephin (approved):
Thank you!

PR #648, opened 2022-05-11
move go CI to github actions

dhiaayachi:
This PR move go checks to a github action workflow and bump go version used in the tests to 1.18.1


.github/workflows/check.yml
```diff
@@ -0,0 +1,77 @@
+name: Checks
+
+on:
+  pull_request:
+
+env:
+  GO_VERSION: "1.18.1"
+
+# This workflow runs for not-yet-reviewed external contributions and so it
```
eculver, line 9: Thank you for adding this note <3

.github/workflows/check.yml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - name: "Fetch source code"
+        uses: actions/checkout@v2
+
+      - name: Install Go toolchain
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      # NOTE: This cache is shared so the following step must always be
```
eculver, line 28: Interesting.. I hadn't seen this be done in other workflows, but good to know.

.github/workflows/check.yml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Fetch source code"
+        uses: actions/checkout@v2
+
+      - name: Install Go toolchain
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
```
eculver, line 56: I'm curious if we should a) match what is currently in CircleCI (Go 1.16) or b) even better, now that we're using GHAs, use a matrix of versions and test against 1.16, 1.17 and 1.18 all at once?

.github/workflows/check.yml
```diff
@@ -0,0 +1,77 @@
...
+          path: "~/go/pkg"
+          key: go-mod-${{ hashFiles('go.sum') }}
+          restore-keys: |
+            go-mod-
+      - name: "Unit tests"
+        run: |
+          go test ./...
+      - name: "Race Unit tests"
+        run: |
+          go test -race ./serf/...
```
eculver, line 43: Can we run these in parallel? Or just run all the tests with the race detector?

.github/workflows/check.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: "go.mod and go.sum consistency check"
+        run: |
+          go mod tidy
+          if [[ -n "$(git status --porcelain)" ]]; then
+            echo >&2 "ERROR: go.mod/go.sum are not up-to-date. Run 'go mod tidy' and then commit the updated files."
+            exit 1
+          fi
+      - name: "go vet"
+        run: |
+          go vet ./...
```
eculver, line 77: I'm curious if we should preserve the existing message for relaying failures? 
https://github.com/hashicorp/serf/pull/648/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L24

eculver (commented):
SOLID. Thanks for doing this. I just have a few small questions.

.github/workflows/check.yml
dhiaayachi: This is not possible the other packages have some races (mostly in test) and we need to fix those before being able to run all the tests with race detector. Yes we can run them in parallel, I will make the change.

.github/workflows/check.yml
dhiaayachi: we are actually missing the gofmt check, I will add it.

serf/snapshot_test.go
```diff
@@ -391,6 +391,7 @@ func TestSnapshotter_leave_rejoin(t *testing.T) {
 }
 
 func TestSnapshotter_slowDiskNotBlockingEventCh(t *testing.T) {
+	t.Skip("Flaky test")
```
dhiaayachi, line 394: @banks This is a test you wrote a while back. It's constantly flaking and I'm wondering if it's a perf regression or an issue with the test environment (it flake in circle as much as GHA). 

In any case, I think this is more a benchmark test and we will probably need to write it using a different approach. 
What do you think? is it safe to disable it for the time being (or even remove it) or this is a regression that we need to spend more time investigating?

.github/workflows/check.yml
```diff
@@ -0,0 +1,113 @@
...
+          go-version: ${{ matrix.GO_VERSION }}
+
+      # NOTE: This cache is shared so the following step must always be
+      # identical across the unit-tests and consistency-checks
+      # jobs, or else weird things could happen.
+      - name: Cache Go modules
+        uses: actions/cache@v3
+        with:
+          path: "~/go/pkg"
+          key: go-mod-${{ hashFiles('go.sum') }}
```
eculver, line 93: Do we need to bake the `${{ matrix.GO_VERSION }}` into our cache key? I'm curious if the mod cache is version-specific 🤔.

dhiaayachi: good question, I need to check this. I would say yes we need to.

serf/snapshot_test.go
banks: 👍 

```
        // I typically see this complete in around
	// 115ms on my machine so this should give plenty of headroom for slower CI
	// environments while still being low enough that actual disk IO would
	// reliably blow it.
```

:D I guess I was wrong then! I think it was a more than just a benchmark in the sense that I wanted to verify that our implementation actually didn't allow disk IO to block gossip and effectively break the cluster but yeah see the commentary in https://github.com/hashicorp/serf/pull/526 - we assumed this was lenient enough for CI but if it's not then disabling seems OK?

.github/workflows/check.yml
dhiaayachi: Added the Go version to the cache key, should be better now.

eculver (approved):
LGTM

PR #649, opened 2022-05-11
fix an out of bound access in `keyListResponseWithCorrectSize`

### LINKED ISSUE #647: response size check on keyListResponseWithCorrectSize breaks with 32bit keys and keyring size >20
The function [fkeyListResponseWithCorrectSize](https://github.com/hashicorp/serf/blob/master/serf/internal_query.go#L163) has an issue in the logic used to decide when to truncate the response. It does a decreasing loop starting from `q.serf.config.QueryResponseSizeLimit / minEncodedKeyLength`, and truncates the list until the message can fit into `q.serf.config.QueryResponseSizeLimit`. If the value of `q.serf.config.QueryResponseSizeLimit / minEncodedKeyLength` higher than `len(resp.Keys)`, the [truncate](https://github.com/hashicorp/serf/blob/master/serf/internal_query.go#L183) operation causes a panic.
We discover the bug in consul, with the default configuration `q.serf.config.QueryResponseSizeLimit / minEncodedKeyLength` is 40, and  it takes 38 16bit keys or 22 32bit keys to trigger the bug and crash a consul server sending a list keyring rpc. The [test](https://github.com/hashicorp/serf/blob/master/serf/internal_query_test.go#L123) doesn't catch the bug as it uses a list with 50 items.

dhiaayachi:
This fix #647

### DIFF
--- a/serf/internal_query.go
+++ b/serf/internal_query.go
@@ -163,6 +163,13 @@ func (s *serfQueries) handleConflict(q *Query) {
 func (s *serfQueries) keyListResponseWithCorrectSize(q *Query, resp *nodeKeyResponse) ([]byte, messageQueryResponse, error) {
 	maxListKeys := q.serf.config.QueryResponseSizeLimit / minEncodedKeyLength
 	actual := len(resp.Keys)
+
+	// if the provided list of keys is smaller then the max allowed, just iterate over it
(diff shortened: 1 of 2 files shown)


markan (approved):
This looks good to me. Curious how often this; it feels like this bug could happen pretty frequently.

PR #653, opened 2022-07-28
go.mod: upgrade to go 1.16

huikang:
- make it consistent with CI job


dhiaayachi:
Hey @huikang ,
The go version in the go mod file represent the minimal (oldest) go version that can be used with the library. 
I'm not sure we have to upgrade it to go 1.18 for serf as we don't use any 1.18 only feature in this library.

huikang:
Hi, @dhiaayachi , thanks for the comment. I hit some issue when compiling the code due to incompatibility issue with `golang.org/x/sys/unix `. So I test changing the go version and upgrade the go.sum for that package.

The new commit has reverted the minimal version to 1.12. Please take another look. Thanks.

PR #657, opened 2022-08-08
Bugfix: fix 'panic: close of closed channel' by atomically swapping the 'closed' member of RPC client handler structs

JHartman5:
We're regularly seeing errors like `panic: close of closed channel` in our application with a stack trace like:
```
goroutine 502383 [running]:
github.com/hashicorp/serf/client.(*queryHandler).Cleanup(0xc00198c060)
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:629 +0x91
github.com/hashicorp/serf/client.(*RPCClient).deregisterHandler(0xc00034c070, 0x11bea00)
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:804 +0xd8
github.com/hashicorp/serf/client.(*queryHandler).Handle(0xc00198c060, 0x11beb00)
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:612 +0x1cb
github.com/hashicorp/serf/client.(*RPCClient).respondSeq(0xc00034c070, 0x11beb00, 0xc001cb3740)
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:824 +0xc4
github.com/hashicorp/serf/client.(*RPCClient).listen(0xc00034c070)
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:840 +0x86
created by github.com/hashicorp/serf/client.ClientFromConfig
	/go/src/github.com/boxcast/playlist_service/vendor/github.com/hashicorp/serf/client/rpc_client.go:148 +0x4bd
```

While I am not 100% sure of the root cause of this, inspecting the RPC client code suggests that there might be a race if someone is calling `client.Close()` on an RPC client at the same time a client might be closing itself due to an error. Since none of the handlers' `Cleanup` methods are currently doing anything atomic, it's easy to see how multiple callers might be `close()`ing the channels at the same time.

This PR attempts to fix the race condition by changing the the `closed` members from `bool` to `uint32`, then rather than checking them and setting them in separate parts of `Cleanup`, using `atomic.CompareAndSwapUint32()` to get and check the old value while also setting it (this should be less overhead than using a mutex).

I noticed that there may be a similar race with the `init` members, so they got the same treatment out of paranoia.

It doesn't seem like there are any tests for the client package, and the top-level tests failed for me before making any changes, so if someone can suggest how I can confirm my changes, I'm more than happy to do so.

### DIFF
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -405,7 +405,7 @@ func (c *RPCClient) GetCoordinate(node string) (*coordinate.Coordinate, error) {
 type monitorHandler struct {
 	client *RPCClient
 	closed uint32 // atomic
-	init   bool
+	init   uint32 // atomic
 	initCh chan<- error
 	logCh  chan<- string
 	seq    uint64
@@ -413,8 +413,7 @@ type monitorHandler struct {
 
 func (mh *monitorHandler) Handle(resp *responseHeader) {
 	// Initialize on the first response
-	if !mh.init {
-		mh.init = true
+	if atomic.CompareAndSwapUint32(&mh.init, 0, 1) {
 		mh.initCh <- strToError(resp.Error)
 		return
 	}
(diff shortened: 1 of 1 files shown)


JHartman5:
I'm going to put my two open PRs (this one and https://github.com/hashicorp/serf/pull/660) together and re-open them, as they're now mutually dependent.

PR #660, opened 2022-08-17
Allow callers to use a custom logger when creating an RPC client

JHartman5:
The `serf.Config` provides the ability to customize the logging by providing your own `*log.Logger`, but the RPC client currently does not. It would be nice to be able to customize the logging from an RPC client as well.


JHartman5:
I'm going to put my two open PRs (this one and https://github.com/hashicorp/serf/pull/657) together and re-open them, as they're now mutually dependent.

Issue #662, opened 2022-09-09
When the nodes leave and start at the same time in a large-capacity consul cluster, some member states will continue to keep leaving，not alive.


yangjunmyfm192085:
I am in a consul scenario with 3 servers and 150 clients. 
When 150 clients leave at the same time and restart, some member states show leaving
Further analysis, the following scenarios appear
- Node A and Node B are both rebooted
- After node A starts, it receives the leave message from B, and A caches the type and ltime of the message to recentIntents
- When A receives the join message from B, the join message is smaller than the leave message ltime. As a result,  In A consul members, the state of B is always leaving.

The source code is as follows

https://github.com/hashicorp/serf/blob/830be12c38e509faa64a2a650629e91c8930a040/serf/serf.go#L1111

https://github.com/hashicorp/serf/blob/830be12c38e509faa64a2a650629e91c8930a040/serf/serf.go#L1220

https://github.com/hashicorp/serf/blob/830be12c38e509faa64a2a650629e91c8930a040/serf/serf.go#L956

https://github.com/hashicorp/serf/blob/830be12c38e509faa64a2a650629e91c8930a040/serf/serf.go#L1225

/kind bug

PR #663, opened 2022-09-21
Fix the bug that consul is always in the leaving state

yangjunmyfm192085:
Fix: https://github.com/hashicorp/serf/issues/662
The function `handleNodeJoin`  is used to set `member.Status = StatusAlive`, and the function `handleNodeJoinIntent` is used to set `member.statusLTime = joinMsg.LTime`.
Unfortunately, the two are not in sync. If set `member.statusLTime` directly in `handleNodeJoinIntent`, will get wrong results. As a result, the member state we see has always been leaving
e.g.
We have 3 members A B and C in our cluster, Both B and C join the cluster through A.
- We do leave and start operations on node B
- When node A updates member.statusLTime of B to the latest value in the function handleNodeJoinIntent, member.Status may still be left. 
- When A sends its member status to C through pushPullMsg, C recognizes that the status of B is leave and marks member .statusLTime is the received value + 1 incorrectly.
- C refreshes the status of member B as leaving, and marks member.statusLTime as member.statusLTime as the received value +1
- The state of B in C continues to be leaving eventually

### DIFF
--- a/serf/serf.go
+++ b/serf/serf.go
@@ -108,6 +108,9 @@ type Serf struct {
 
 	// metricLabels is the slice of labels to put on all emitted metrics
 	metricLabels []metrics.Label
+
+	// lamport clock time of last received join intent message
+	memberJoinIntentstatusLTime map[string]LamportTime
 }
 
 // SerfState is the state of the Serf instance.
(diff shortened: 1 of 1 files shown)


yangjunmyfm192085 (commented):
/cc @rboyer

PR #685, opened 2023-04-16
Switch logging to hclog

sinkingpoint:
Currently, serf takes a standard lib log.Logger. This makes using serf as a library a bit weird because you have to have a logger that may produce output in a different format to your main binary. There are ways to hack around this, e.g. by using a log.Logger that writes to some intermediate buffer that parses the log output, but this loses a lot of context that we would otherwise have.

In this commit, we swap serf over to taking an hclog.Logger instead. As an interface, this means that library consumers can use whatever logger they want, and we can use hclog's built-in support for log levels and structured logging.


freddygv (changes requested):
Hi @sinkingpoint, thanks for the contribution, this is heading in a good direction. 

The main thing I noticed is that the `command/agent` pkg still uses a `*log.Logger` in various ways, which leads to mismatched logs when running the Serf binary.  

Notice the timestamp differences below:

> ==> Log data will now stream in as it occurs:
> 
>     2023/04/17 11:05:12 [INFO] agent: Serf agent starting
>     2023-04-17T11:05:12.814-0600 [INFO]  serf: EventMemberJoin: test 10.0.0.145
>     2023/04/17 11:05:13 [INFO] agent: Received event: member-join
>     2023/04/17 12:07:24 [DEBUG] memberlist: Stream connection from=[::1]:54771

Could you update this to make the logging consistent for events coming from the serf library, the command/agent package, and memberlist?

sinkingpoint:
Thanks for the review @freddygv ! I was being a bit lazy and only changed the path I was using 😂. I've just pushed a change that I think updates the rest of the code. The only thing I'm a bit unsure on is if we want to stop passing the `logOutput` around and instead pass around the complete hclog.Logger so that subcomponents don't have to make their own.

freddygv:
> The only thing I'm a bit unsure on is if we want to stop passing the logOutput around and instead pass around the complete hclog.Logger so that subcomponents don't have to make their own.

Hmm I'm not sure about that either at the moment. I'll do some thinking about it.

PR #695, opened 2023-07-18
disable udp6 on windows

phanirithvij:
- disable udp6 on windows (after disabling serf works on my windows10 device with -discover flag over lan)
- update mdns to 1.0.5 (DisableIPv6 was introduced in this version)

See https://github.com/hashicorp/mdns/issues/80


phanirithvij:
Can you take a look at this @freddygv? thanks.

phanirithvij:
I no longer use windows and haven't used serf much.

PR #705, opened 2023-10-19
Upgrade to msgpack v2 2.1.2

swenson:
This adds the option to specify what version of the `time.Time` encoding format is desired. The default option is `false`, which preserves compatibility with the current dependency of
`hashicorp/go-msgpack 0.5.5` in go.mod, but users of this library will probably want to override the the setting.

While we're here, upgrade the go.mod file.

I tested that this appears to work with Nomad.

This needs to be updated and merged after
https://github.com/hashicorp/memberlist/pull/292


client/rpc_client.go
```diff
@@ -53,6 +53,12 @@ type Config struct {
 	// If provided, overrides the DefaultTimeout used for
 	// IO deadlines
 	Timeout time.Duration
+
+	// MsgpackUseNewTimeFormat when set to true, force the underlying msgpack
+	// codec to use the new format of time.Time when encoding (used in
+	// go-msgpack v1.1.5 by default). Decoding is not affected, as all
+	// go-msgpack v2.1.0+ decoders know how to decode both formats.
```
dhiaayachi, line 60: Same comment I left about the godoc in the memberlist PR.

client/rpc_client.go
```diff
@@ -144,10 +150,14 @@ func ClientFromConfig(c *Config) (*RPCClient, error) {
...
-	client.enc = codec.NewEncoder(client.writer,
-		&codec.MsgpackHandle{RawToString: true, WriteExt: true})
+	handle := &codec.MsgpackHandle{
+		WriteExt: true,
+		BasicHandle: codec.BasicHandle{
+			TimeNotBuiltin: !c.MsgpackUseNewTimeFormat,
+		},
+	}
+	client.dec = codec.NewDecoder(client.reader, handle)
+	client.enc = codec.NewEncoder(client.writer, handle)
```
dhiaayachi, line 160: I think it's safer to create 2 separate `handle` for the encoder and the decoder. While I don't think it's an issue now as we don't keep any state in the handle, but I think it introduces a possible foot gun if we do so in the future.

cmd/serf/command/agent/ipc.go
```diff
@@ -393,10 +395,14 @@ func (i *AgentIPC) listen() {
...
-		client.enc = codec.NewEncoder(client.writer,
-			&codec.MsgpackHandle{RawToString: true, WriteExt: true})
+		handle := &codec.MsgpackHandle{
+			WriteExt: true,
+			BasicHandle: codec.BasicHandle{
+				TimeNotBuiltin: !i.msgpackUseNewTimeFormat,
+			},
+		}
+		client.dec = codec.NewDecoder(client.reader, handle)
+		client.enc = codec.NewEncoder(client.writer, handle)
```
dhiaayachi, line 405: same comment as above here about handle

serf/messages_test.go
```diff
@@ -23,7 +23,7 @@ func TestQueryFlags(t *testing.T) {
 
 func TestEncodeMessage(t *testing.T) {
 	in := &messageLeave{Node: "foo"}
-	raw, err := encodeMessage(messageLeaveType, in)
+	raw, err := encodeMessage(messageLeaveType, in, false)
```
dhiaayachi, line 26: Maybe we can add some tests here that encode/decode with both options a message that have a time field in it 🤔. Not sure, as I guess all of this is already tested in the library. WYT?

dhiaayachi (commented):
Thank you @swenson! I left few comments for your consideration.

serf/messages_test.go
swenson: I test this pretty thoroughly in the library https://github.com/hashicorp/go-msgpack/blob/main/codec/raft_test.go and in the Raft library https://github.com/hashicorp/raft/blob/main/util_test.go#L15

dhiaayachi (approved):
🚀

swenson:
Thanks!

PR #720, opened 2024-03-19
serf/snapshot.go: replace os.SEEK with io.Seek

shubhisroking:
ref: https://pkg.go.dev/os#SEEK_SET


tgross:
Pulled into https://github.com/hashicorp/serf/pull/819

PR #733, opened 2024-07-02
Enhance mDNS Functionality

badrabubker:
This PR:
* Allows specifying `iface` for mDNS: In our use case, we disable multicast on the interface that Serf binds to. However, we have another network where Serf can discover neighbors using multicast.
* Updates mDNS to v1.0.5: mDNS v1.0.5 enables the disabling of one of the IP stacks.
* Allows disabling one of the IP stacks in the config file and CLI arguments.

### DIFF
--- a/cmd/serf/command/agent/command.go
+++ b/cmd/serf/command/agent/command.go
@@ -89,6 +89,8 @@ func (c *Command) readConfig() *Config {
 	cmdFlags.StringVar(&cmdConfig.Discover, "discover", "", "mDNS discovery name")
 	cmdFlags.StringVar(&cmdConfig.Interface, "iface", "", "interface to bind to")
 	cmdFlags.StringVar(&cmdConfig.MDNS.Interface, "mdns-iface", "", "interface to use for mDNS")
+	cmdFlags.BoolVar(&cmdConfig.MDNS.DisableIPv4, "mdns-disable-ipv4", false, "disable IPv4 for mDNS")
+	cmdFlags.BoolVar(&cmdConfig.MDNS.DisableIPv6, "mdns-disable-ipv6", false, "disable IPv6 for mDNS")
 	cmdFlags.StringVar(&cmdConfig.TagsFile, "tags-file", "", "tag persistence file")
 	cmdFlags.BoolVar(&cmdConfig.EnableSyslog, "syslog", false,
 		"enable logging to syslog facility")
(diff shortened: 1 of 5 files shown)


cmd/serf/command/agent/config.go
```diff
@@ -154,10 +162,12 @@ type Config struct {
 	// allows Serf agents to join each other with zero configuration.
 	Discover string `mapstructure:"discover"`
 
+	MDNS MDNSConfig `mapstructure:"mdns"`
+
 	// Interface is used to provide a binding interface to use. It can be
 	// used instead of providing a bind address, as Serf will discover the
 	// address of the provided interface. It is also used to set the multicast
-	// device used with `-discover`.
+	// device used with `-discover`, if mdns.interface is not set
```
DanStough, line 170: nit:
```suggestion
	// device used with `-discover`, if `mdns-iface` is not set
```

cmd/serf/command/agent/command.go
```diff
@@ -734,7 +752,10 @@ Options:
                            -bind if the interface is known but not the address.
                            If both are provided, then Serf verifies that the
                            interface has the bind address that is provided. This
-                           flag also sets the multicast device used for -discover.
+                           flag also sets the multicast device used for -discover,
+                           if mdns-iface is not specified.
+  -mdns-iface              Network interface to use for mDNS. If not provided, the
+                           -iface value is used.
```
DanStough, line 758: Need doc for the other boolean flags too.

cmd/serf/command/agent/command.go
```diff
@@ -176,6 +179,19 @@ func (c *Command) readConfig() *Config {
 		return nil
 	}
 
+	if config.MDNS.Interface != "" {
```
DanStough, line 182: We should also handle the error case where `mdns.interface` is specified w/o `discovery`.

DanStough (changes requested):
Thank you for the PR! I had some really minor comments we can address before merging.

cmd/serf/command/agent/command.go
badrabubker: Added

cmd/serf/command/agent/command.go
badrabubker: I added a check but I was not sure whether we should also add a check for the disable-ip-v(4|6) flags

cmd/serf/command/agent/config.go
badrabubker: Thanks a lot for your review. I added your suggestion in the amended commit

cmd/serf/command/agent/command.go
DanStough: I don't think we need it, because it is valid to inherit the top-level `iface` in that case. 👍

DanStough (approved):
LGTM 🚀 . I'll wait for the checks to pass before merging.

Issue #18, opened 2013-10-23
Join/Leave event flushing causes active agents to show as 'leaving'


darkone23:
To replicate:

start foo
`$ serf agent -node=foo -bind=127.0.0.10 -rpc-addr=127.0.0.1:7373`

start bar
`$ serf agent -node=bar -bind=127.0.0.11 -rpc-addr=127.0.0.1:7374`

tell foo to join bar
`$ serf join -rpc-addr=127.0.0.1:7373 127.0.0.11`

at this point, both show the other as alive, now control-c foo

bar now sees foo as 'left'
`$ serf members -rpc-addr=127.0.0.1:7374`

```
bar    127.0.0.11    alive
foo    127.0.0.10    left
```

restart foo:
`$ serf agent -node=foo -bind=127.0.0.10 -rpc-addr=127.0.0.1:7373`

both don't see the other, as expected -- now tell bar to join foo:
`$ serf join -rpc-addr=127.0.0.1:7374 127.0.0.10`

In this state, bar sees that 'foo' is alive, but foo thinks that 'foo' is leaving.
`$ serf members`

```
foo    127.0.0.10    leaving
bar    127.0.0.11    alive
```

PR #738, opened 2024-08-20
s2 should see its status as alive after fixing issue #18

mbver:
the test doesn't reflect the fix of issue #18. before apply fixing, a node rejoins the cluster still sees itself in "leaving" state while other nodes see it as alive. now, it sees itself as alive.

### DIFF
--- a/serf/serf_test.go
+++ b/serf/serf_test.go
@@ -1699,17 +1699,17 @@ func TestSerf_joinLeaveJoin(t *testing.T) {
 			r.Fatalf("s2 members: %d", s2.NumNodes())
 		}
 
-		// s1 should see the node as alive
-		mems := s1.Members()
-		anyLeft := false
+		// s2 should see the itself as alive
+		mems := s2.Members()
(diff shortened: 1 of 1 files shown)


skpratt:
Hi @mbver, if you could update the description to add more background detail for the expected state of s2's members, so it is clear what this test is covering, that would be great.
It also looks like the test as written in master is not currently failing. Could you please add a description for why the existing test to verify the state of s1's members should be removed? Rather than adding a separate check for s2?

skpratt (changes requested):
Hi @mbver, if you could update the description to add more background detail for the expected state of s2's members, so it is clear what this test is covering, that would be great.
It also looks like the test as written in master is not currently failing. Could you please add a description for why the existing test to verify the state of s1's members should be removed? Rather than adding a separate check for s2?

mbver:
thanks @skpratt . test for s1 is added.

skpratt (approved):

PR #741, opened 2024-08-21
guard against nil iface

mbver:


tgross:
Sorry, didn't see that this was the root cause the issue fixed by https://github.com/hashicorp/serf/pull/817

PR #750, opened 2024-10-17
remove deprecated ioutil.

tristanmorgan:
Replace usage of deprecated `io/ioutil` with functionally equivalent usage of `os` and `io` packages.


tgross:
Pulled into https://github.com/hashicorp/serf/pull/819

PR #760, opened 2024-12-18
Introduce hashicorp/go-metrics compatibility

mkeeler:
This PR migrates the library to using `github.com/hashicorp/go-metrics/compat` instead of `github.com/armon/go-metrics`.

Applications consuming the library can control whether metrics get emitted to either `armon/go-metrics` or `hashicorp/go-metrics` by specifying either the `armonmetrics` or `hashicorpmetrics` build tags. The default behavior is currently to prefer `armon/go-metrics` for maximum backwards compatibility. 

Future PRs will:
* Bump the `hashicorp/go-metrics` and `hashicorp/memberlist` versions to one where the default behavior is to emit metrics using `hashicorp/go-metrics` (Timeline: mid-2025)
* Remove `compat` package usage with direct usage of `hashicorp/go-metrics` (Timeline: end-2025)

TODO:
- [x]  Pull in a tagged version of `hashicorp/go-metrics` once [upstream PR](https://github.com/hashicorp/go-metrics/pull/169) is merged
- [x] Pull in a tagged version of `hashicorp/memberlist` once [upstreamPR](https://github.com/hashicorp/memberlist/pull/315) is merged


mukeshjc (approved):
LGTM

RSWilli:
How is it that `armon/go-metrics` is only deprecated, when you can't even install it from a fresh go mod cache? If you click on the URL: https://github.com/armon/go-metrics then you're redirected to https://github.com/hashicorp/go-metrics, which the go compiler doesn't like.

mkeeler:
Maybe I should clarify the wording in the readme. 

This library (and all others within the `hashicorp` namespace) emitting metrics to the `armon/go-metrics` global handler is deprecated. Right now the compatibility shim in place allows metrics to flow to either of the modules but that isn't a state we want to maintain indefinitely. Mid 2025 my plan is to flip the default behavior to prefer emitting metrics towards `hashicorp/go-metrics` (build tags can still opt-out). Then at the end of the year my plan is to remove the compatibility layer all together and only emit metrics to `hashicorp/go-metrics`.

As for using `armon/go-metrics`, you can still `go get github.com/armon/go-metrics@v0.4.1`. When Go finds this tag, the module declared in the downloaded go.mod will be `armon/go-metrics` and the compiler wont complain. Using, `latest` or any of the later versions is broken due to the module renaming.

PR #771, opened 2025-03-10
chore: replace github.com/mitchellh/mapstructure with github.com/go-viper/mapstructure/v2

hazzik:


ghost (commented):
Thanks for raising the PR but not exactly sure why we need this change. Were you getting blocked on anything due to the original dependency ?

hazzik:
Not blocked. Just do not like having outdated dependencies. Original dependency was deprecated by the owner, and the new one is a "blessed fork" that will be supported by viper team. And serf seems to be a low level dependency in quite a lot of packages that we use. So my motivation was to have the train moved in that direction.

thaJeztah:
cc @rmainwork are you a maintainer for this repo? Was curious as well if this project would update to the maintained / blessed fork (I'm trying to prevent dependency sprawl in my projects)

tgross:
Thanks for the PR @hazzik and apologies for the long delay on getting this merged as well. I've rebased on tip so I'll need to get a colleague to approve and then we can finally land this. (Failing tests are associated with https://github.com/hashicorp/serf/issues/813)

jrasell (approved):

PR #778, opened 2025-09-24
build: update Go to 1.24

pkazmierczak:
## Description

This PR updates Go build requirement to 1.24 in order to upgrade transitive dependencies like golang.org/x/tools to newer versions that would compile with recent Go (0.14.0 does not). 

## Related Issue

https://github.com/hashicorp/nomad/pull/26823

## How Has This Been Tested?

Via unit tests.


pkazmierczak:
I did some more testing on this. Setting go toolchains to different versions from 1.19.13 to 1.25.1 and running with go.mod as it is on `master` makes all the performance tests pass. So the issue is with one of the dependencies. Once I upgrade `go-msgpack` to 2.1.5 and `golang.org/x/tools` to 0.37, they start to fail roughly 70% of the time on my local laptop. 

Once again, I'd appreciate some insight into how these performance tests work and what do they measure exactly, and most importantly where do [these numbers](https://github.com/hashicorp/serf/blob/dcd313de5ff4150661734bb417bc100e9f238ad3/coordinate/performance_test.go#L23) come from. I'm happy to fix the tests, but I need some guidance.

coordinate/phantom.go
```diff
@@ -145,10 +145,10 @@ func GenerateRandom(nodes int, mean time.Duration, deviation time.Duration) [][]
 // underlying algorithm which will use random numbers for position vectors when
 // starting out with everything at the origin).
 func Simulate(clients []*Client, truth [][]time.Duration, cycles int) {
-	rand.Seed(1)
+	rand.New(rand.NewSource(1))
 
 	nodes := len(clients)
-	for cycle := 0; cycle < cycles; cycle++ {
+	for range cycles {
```
CreatorHead, line 151: I was going through the diff, and noticed that along with the Go/toolchain + dependency bump, there are also changes in coordinate/phantom.go, e.g. new calls to `rand.New(rand.NewSource(1))` and an extra loop (for range cycles).

Could you clarify why these were introduced? From the PR title/description it looks like the intent was mainly a version bump, so I wanted to understand if these RNG/loop changes are intentional, or if they might be leftovers from local testing.

schmichael:
The performance test failures seem to be related to Go 1.24 making `rand.Seed(...)` a noop, but the performance tests relied on its Go 1.16-esque behavior. I added an env var to revert to the old behavior just for the tests, and that seems to have fixed things.

sonamtenzin2 (approved):
Thanks for looking into this 🚀

CreatorHead (approved):

tgross:
Closed by https://github.com/hashicorp/serf/pull/810

PR #786, opened 2026-01-16
Fix ci checks

ssagarverma:
## Description
Update Go version matrix in CI workflow to fix CI checks that are stopping various other PRs.

Note: This branch is based on #768 and is intended to verify that the fix works correctly with Dependabot upgrades, without introducing noise into #768. Most of the changes here are carried over from that PR.

## Related Issue
#768 

## How Has This Been Tested?
`golangci-lint run ./...`


jrasell (commented):
Hi @ssagarverma, can we add a description of what CI bugs or problems this is fixing please? It also looks like this is updating memberlist, msgpack, and the Go Module tooling version, which doesn't feel right given the title.

ssagarverma:
> Hi @ssagarverma, can we add a description of what CI bugs or problems this is fixing please? It also looks like this is updating memberlist, msgpack, and the Go Module tooling version, which doesn't feel right given the title.

@jrasell I've updated the intention of this PR in the description. Although I'm still seeing a couple of unit test failures (TestPerformance_Grid, TestPerformance_Height).

tgross:
Superceded by https://github.com/hashicorp/serf/pull/810

PR #794, opened 2026-01-23
Changed go version to 1.24

sonamtenzin2:
## Description
Needs to be changed to 1.24 in order to support dependency upgrades which fix vulnerabilities
Added retries in performance_test.go because they keep failing due to randomness


tgross:
Closed by https://github.com/hashicorp/serf/pull/810

PR #803, opened 2026-04-05
feat(metrics): add exposing metrics functionality

badrabubker:
<!-- heimdall_github_prtemplate:grc-pci_dss-2024-01-05 -->
## Description
 **Background**                                                                                                     
                                                                                                                 
  I use Serf to monitor the reachability of a network cluster that is not routed outside the cluster. I need to  
  expose metrics describing the state of each node so an external Prometheus instance can scrape them.
                                                                                                                 
  **Why not expose from every node?**                                                                                                                       
   
  Consistency: while gossip is converging after a membership change, different nodes may temporarily disagree on
   the state of a peer. Scraping all nodes during this window would produce conflicting values for the same    
  series in Prometheus, making alerts unreliable.                                                                
                                                                                                               
  Cardinality: if every node reports the status of every other node, the number of time series grows as O(n²).  
  For a cluster of 100 nodes that is 10,000 series for a single metric. Electing one exporter keeps it O(n).
                                                                                                                 
  Solution: elected metrics exporter                                                                             
   
  Only one node at a time holds the tag role=metrics_exporter and serves serf_node_status for the whole cluster. 
  All nodes always expose serf_event_received and serf_metrics_exporter (0 or 1) regardless of their role.     
                                                                                                                 
  **Election algorithm**                                                                                             
   
  The election is intentionally simple and deterministic, no coordination protocol needed because SWIM already  
  provides an eventually-consistent view of membership:                                                        
                                                                                                                 
  1. On any membership event (join, leave, failed, update, reap), each node re-evaluates the exporter role.      
  2. If a node sees fewer than 2 alive peers it considers itself potentially isolated (split-brain) and steps
  down rather than claiming to speak for the cluster.                                                            
  3. If there is no current exporter and at least 2 alive nodes are visible, the alphabetically first alive node
  elects itself by setting role=metrics_exporter via SetTags, which gossips the change to the rest of the        
  cluster.                                                                                                     
  4. If gossip convergence temporarily produces multiple exporters, all but the alphabetically first one step    
  down on the next membership event.                                                                             
  5. Tags are only updated when the role actually changes, to avoid unnecessary gossip churn on every event.


## Related Issue

<!-- If this PR is linked to any issue, provide the issue number or description here. Any related JIRA tickets can also be added here. -->

## How Has This Been Tested?

 started 3 agents with --metrics enabled on different ports, joined them into a    
  cluster, verified that:                                                                                      
                                                                                                                 
  - only node1 (alphabetically first) got elected as exporter and served serf_node_status for all 3 nodes        
  - node2 and node3 showed serf_metrics_exporter 0 with no node status series                                    
                                                                                                                 
  Then killed node1, force-left it, and confirmed node2 re-elected itself and started serving the correct status 
  including serf_node_status{node="node1"} 3 (left).


badrabubker:
I bumped go to 1.21 to use `slices` but I see now most checks have failed. I can downgrade to 1.19 later if the team decided to go with the PR

tgross (changes requested):
Sorry for the delay on review here, but serf the application (as opposed to the library) is effectively in maintenance mode at best at this point. An election algorithm for metrics is not something we'd likely want in any case.

That being said, the serf agent code base already has `go-metrics`. If we're going to expose Prometheus metrics from the agent, it should be adding a `go-metric` sink to https://github.com/hashicorp/serf/blob/a24f620f9e5b291f6d4610e275086250463eeb64/cmd/serf/command/agent/command.go#L577-L614 rather than hand-rolling a new exporter.

PR #807, opened 2026-05-12
Add validation and warning for force-leave command

### LINKED ISSUE #500: force-leave is not validating input

kimjune01:
The force-leave command did not validate whether the specified node exists in the cluster, giving no visual indication when operating on a non-member. This made it difficult for operators to distinguish between successful operations and no-ops.

## What Changed

Query the member list before broadcasting force-leave. If the node is not found, emit a warning. If the member query fails, emit a warning about the failure but continue with force-leave (maintaining idempotent behavior with exit code 0).

This provides better observability in normal operations while remaining robust during degraded network conditions.

Adds test coverage for the nonexistent node scenario.

Fixes #500

### DIFF
--- a/cmd/serf/command/force_leave.go
+++ b/cmd/serf/command/force_leave.go
@@ -47,7 +47,9 @@ func (c *ForceLeaveCommand) Run(args []string) int {
 	defer client.Close()
 
 	members, err := client.Members()
-	if err == nil {
+	if err != nil {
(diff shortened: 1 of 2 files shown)


kimjune01:
Cannot validate on current setup, so drafting. Please close or take it over.

PR #816, opened 2026-06-10
fix coordinate performance tests

### LINKED ISSUE #813: test flakes: TestPerformance_Height and TestPerformance_Grid
These two tests have become flaky at some point. Ex https://github.com/hashicorp/serf/actions/runs/27279725767/job/80570553373?pr=793 I've got a body of work to do on this library to get our CI current and make sure dependencies are getting updated properly. Once that work is done we can address these.

tgross:
In Go 1.24+ attempting seeding the global random number generator is a no-op. The tests for the coordinate package relied on this to get deterministic configuration of the "random" clusters they're simulating. Without a deterministic RNG tests become flaky. But once this is fixed, a few flaky tests were now failing deterministically. This appears to because the thresholds were too tight in the face of small changes in the toolchain.

Update the test configuration to allow using a non-global RNG, and update the test thresholds to bring the assertions inline with the current behavior.

Fixes: https://github.com/hashicorp/serf/issues/813

### DIFF
--- a/coordinate/config.go
+++ b/coordinate/config.go
@@ -4,7 +4,9 @@
 package coordinate
 
 import (
-	"github.com/hashicorp/go-metrics/compat"
+	"math/rand"
+
+	metrics "github.com/hashicorp/go-metrics/compat"
 )
 
 // Config is used to set the parameters of the Vivaldi-based coordinate mapping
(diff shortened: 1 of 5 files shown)


gulducat (approved):
LGTM!

PR #818, opened 2026-06-10
CI: run checks on push and reduce redundant config

tgross:
Checks were not running on merges to master, which meant that it was harder to track down when regressions were introduced because commits may be merged out of order. Run on every push to master, as we do for other repos.

Also the CI configuration had a bunch of extraneous configuration that was needed for older versions of `setup-go` that's all covered by the action itself now.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -55,12 +55,19 @@ tools:
 	fi
 
 vet:
-	@echo "--> Running go vet"
-	@go vet -tags '$(GOTAGS)' $(GOFILES); if [ $$? -eq 1 ]; then \
+	go vet -tags '$(GOTAGS)' $(GOFILES); if [ $$? -eq 1 ]; then \
 		echo ""; \
 		echo "Vet found suspicious constructs. Please check the reported constructs"; \
 		echo "and fix them if necessary before submitting the code for review."; \
 		exit 1; \
 	fi
 
-.PHONY: default bin cov format dev dist get-tools subnet test testrace tools vet copywriteheaders
+tidy:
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -55,12 +55,19 @@ tools:
 	fi
 
 vet:
-	@echo "--> Running go vet"
-	@go vet -tags '$(GOTAGS)' $(GOFILES); if [ $$? -eq 1 ]; then \
+	go vet -tags '$(GOTAGS)' $(GOFILES); if [ $$? -eq 1 ]; then \
```
gulducat, line 58: We update `make vet` here, but then no workflows run it, and running `go vet` explicitly was dropped from the workflow (as was `go fmt` check, incidentally).

It _would_ be run by `make test` or `make testrace`, but the workflow doesn't use those, either, but rather runs `go test` directly.

Am I missing some transitive dependency or something? This overall arrangement between makefile and workflow is oddly confusing considering how simple it actually wants to be.

tgross: In CI `go vet`/`go fmt` are both covered by `golangci-lint`. 

So I think the thought process here is that you can run `make test` locally and get a good smoke test without having to have all the tools set up. Right now there's nothing in the makefile to get all the tooling setup except for `gox`, which is a dubious requirement. But I'll need to make another pass to get the build of the binaries working without that. So this is an intermediate step.

gulducat (approved):
As an intermediate step acknowledging the slight wonkiness between the makefile and workflow, LGTM!

PR #823, opened 2026-06-12
dependencies: update mdns to 1.0.7

tgross:
The `hashicorp/mdns` library pulled in a lot of stale transitive dependencies that weren't actual vulnerabilities but have been tripping scanners. Pull in the updated version of the library that upgrades a bunch of these deps.

Should clear remaining spurious CodeQL results: https://github.com/hashicorp/serf/security/code-scanning


serf/serf_test.go
```diff
@@ -2027,9 +2027,7 @@ func TestSerf_Query(t *testing.T) {
 
 	ctx := t.Context()
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
```
tgross, line 2030: Not sure why this lint trips now and not on master, but 🤷

gulducat (approved):
👍

PR #831, opened 2026-07-23
handle pruning asynchronously

tgross:
If we get a message to leave with the prune flag set, the handler sleeps for the broadcast timeout + leave propogation delay before erasing the node, in order to give the message time to propogate. But we're holding the `memberLock` mutex while doing so, preventing the node from handling most other packets. Fire off a goroutine to handle the pruning.

Ref: https://hashicorp.atlassian.net/browse/NMD-1625
Ref: https://hashicorp.atlassian.net/browse/PSA-2905


jrasell (approved):
LGTM, thanks @tgross!

PR #832, opened 2026-07-23
bounds-check message field payloads prior to indexing/slicing them

tgross:
The `messageQueryType` message has a `Filters` slice that can potentially be empty, and when we index into that empty slice we'll panic. Similarly, the handlers for the various keyring messages are missing bounds checks on the `Payload` fields, which results in a panic when the slice an empty slice.

Ref: https://hashicorp.atlassian.net/browse/PSA-2905
Ref: https://hashicorp.atlassian.net/browse/NMD-1625


chrisroberts (approved):

jrasell (approved):
LGTM!

